PR #2, opened 2022-11-02
Fix XDP redirect in kind

astoycos:
Try and ensure XDP redirect will work in KIND
  
  This Updates the XDP program to rewrite:
  - Src Mac address
  - Dst Port
  
  and allows us to configure if we want to run cksum
  calculation on a per backend basis. It also fixes
  the existing CKSUM calculation function to ignore
  trailing bits.
  
  Lastly We should be using the autogenerated bpf2go
  map structs from `bpf_*.go`

For running in KIND make sure to turn ON GRO manually, and if using cksums (I wouldn't) turn off CKSUM offloading

### DIFF
--- a/dataplane/README.md
+++ b/dataplane/README.md
@@ -1,0 +1,191 @@
+# Some helpful hints for debugging this XDP program
+
+## Tracing XDP redirect (on first interface where main XDP program is attached)
+
+(TODO finish tracing the XDP path through the kernel)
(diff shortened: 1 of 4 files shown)


shaneutt (approved):
Awesome, thank you Andrew. Just gonna merge this as-is for now since we're in early POC state. I'll do some testing to make sure things are working and then I'll wire up `UDPRoute` to it :+1:

PR #19, opened 2022-11-11
listener status update

### LINKED ISSUE #13: Address and Listener Status in Gateways
## Problem Statement

When we first built the POC we didn't bother much with statuses for objects, we just focused on the high level demo bits. As such we still need to properly implement the address and listener status for `Gateway` resources.

## Acceptance Criteria

- [ ] `Gateway` resource status addresses and listener status should be updated properly according to Gateway API specification
- [ ] an integration test should be added that verifies the status for multiple `Gateway` resources with different configurations

mlavacca:
Initial scaffolding to enforce the Gateway listener status.
Fixes #13


controllers/gateway_controller.go
```diff
@@ -99,37 +96,63 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
+	gatewayReadyStatus, gatewayReadyStatusIsSet := isGatewayReady(gw)
+	oldGateway := gw.DeepCopy()
+	initGatewayStatus(gw)
+	factorizeStatus(gw, oldGateway)
```
shaneutt, line 102: I'm not seeing a need to do this, especially if you take the suggestion above, can we just drop these? Let me know if I'm missing something :thinking: 

```suggestion
```

controllers/gateway_controller.go
```diff
@@ -99,37 +96,63 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+
 	log.Info("checking for Service for Gateway")
 	svc, err := r.getServiceForGateway(ctx, gw)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 	if svc == nil {
+		// if the ready status is not set, or the gateway is marked as ready, mark it as not ready
+		if !gatewayReadyStatusIsSet || gatewayReadyStatus {
+			return ctrl.Result{}, r.patchGatewayStatus(ctx, gw, oldGateway) // status patch will requeue gateway
```
shaneutt, line 112: nit: I'm not sure we really need this `patchGatewayStatus` wrapper, we're trading one line with slightly more characters for one line with slightly less, but I feel readers will often feel compelled to read the implementation anyway to know what it's doing, so I kinda feel like we should just do the `Patch()` here :thinking:

controllers/gateway_controller_test.go
```diff
@@ -90,3 +101,348 @@ func TestGatewayReconciler_gatewayHasMatchingGatewayClass(t *testing.T) {
...
+
+			fakeClient := fakectrlruntimeclient.
+				NewClientBuilder().
+				WithScheme(scheme.Scheme).
+				WithObjects(objectsToAdd...).
+				Build()
+
+			reconciler := GatewayReconciler{
+				Client: fakeClient,
+			}
```
shaneutt, line 443: I'm glad to see us adding some tests for the gateway reconciler, thanks :+1:

controllers/gateway_controller_test.go
```diff
@@ -90,3 +101,348 @@ func TestGatewayReconciler_gatewayHasMatchingGatewayClass(t *testing.T) {
...
+
+			tc.testBody(t, reconciler, tc.gatewayReq, tc.gateway)
```
shaneutt, line 445: nit: I would rename `testBody` to `run` personally:

```suggestion
			tc.run(t, reconciler, tc.gatewayReq, tc.gateway)
```

controllers/gateway_controller.go
```diff
@@ -99,37 +96,63 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
+	gatewayReadyStatus, gatewayReadyStatusIsSet := isGatewayReady(gw)
```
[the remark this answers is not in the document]
mlavacca: I initialized the status as you suggested, and now the `gatewayReadyStatusIsSet` var is checked only at the beginning, and if not set, the status is initialized

controllers/gateway_controller.go
mlavacca: done :)

controllers/gateway_controller.go
mlavacca: they are needed to initialize the status and to avoid useless transitions (patch is effective for conditions which differ in the `transitionTIme`, and we don't want this behavior)

shaneutt (approved):
While I think we could probably tweak this further, I think this is in good shape for where we're at with the project as we're bound to have lots more changes in this area soon since we're so early in the project.

Thank you, LGTM!

Issue #17, opened 2022-11-11
Filter out Gateways before reconciliation


mlavacca:
## Problem Statement

Every time a new `Gateway` gets reconciled by the Gateway controller, the controller must check the:
* `GatewayClass` referenced by the `Gateway` exists
* `GatewayClass.Spec.ControllerName` matches with Blixt's `controllerName` 

These two checks should be moved to a predicate to save reconciliation loops not needed for `Gateway`s not backed by Blixt's `GatewayClass`es.

## Acceptance Criteria
- [ ] the aforementioned checks are moved to a predicate that runs in the Gateway controller.
- [ ] a unit test for such a predicate must be added

Rahul-D78:
Hi @mlavacca can i work on this one

mlavacca:
Hi @Rahul-D78, of course. Feel free to assign yourself 👍

mlavacca:
> /assign

We don't have this kind of automation yet :)

Rahul-D78:
Can we use a constant value for `Blixt's controllerName` or it is predefined.
Can you please give some context regarding this -- `GatewayClass referenced by the Gateway exists`

mlavacca:
> Can we use a constant value for `Blixt's controllerName` or it is predefined. Can you please give some context regarding this -- `GatewayClass referenced by the Gateway exists`

Basically the following portion of code should be put into a predicate:
https://github.com/Kong/blixt/blob/69655b46bd244cbcc48695ad8d3d06d823091e7a/controllers/gateway_controller.go#L63-L73

shaneutt:
Relevant to https://github.com/Kong/blixt/pull/20#pullrequestreview-1177793685 I'm going to add an initial unit test here.

PR #20, opened 2022-11-11
feat: filter out Gateways before reconciliation

Rahul-D78:
Filter out Gateways before reconciliation
Fixes: #17

### DIFF
--- a/controllers/gateway_controller.go
+++ b/controllers/gateway_controller.go
@@ -62,10 +62,12 @@ func (r *GatewayReconciler) gatewayHasMatchingGatewayClass(obj client.Object) bo
 	}
 	gatewayClass := &gatewayv1beta1.GatewayClass{}
 	if err := r.Client.Get(context.Background(), client.ObjectKey{Name: string(gateway.Spec.GatewayClassName)}, gatewayClass); err != nil {
-		r.Log.Error(err, "could not retrieve gatewayclass", "gatewayclass", gateway.Spec.GatewayClassName)
+		if errors.IsNotFound(err) {
+			r.Log.Error(err, "gatewayclass not found", "gatewayclass", gateway.Spec.GatewayClassName)
+		}
 		return false
 	}
-	return gatewayClass.Spec.ControllerName != GatewayClassControllerName
+	return true // some other problem retrieving the object, enqueue anyway to avoid dropping
 }
 
 func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
--- a/controllers/gateway_controller.go
+++ b/controllers/gateway_controller.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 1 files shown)


controllers/gateway_controller.go
```diff
@@ -36,11 +38,25 @@ const gatewayServiceLabel = "konghq.com/owned-by-gateway"
 type GatewayReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
+	Log    logr.Logger
 }
 
 func (r *GatewayReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&gatewayv1beta1.Gateway{}).
+		WithEventFilter(predicate.NewPredicateFuncs(func(obj client.Object) bool {
```
mlavacca, line 47: You should put this predicate as the variadic argument in the `For(&gatewayv1beta1.Gateway{})`, because if we set it globally here, it will be applied to all other resources that will be watched/owned by this reconcile in the future.

mlavacca: Also, please define this `func(obj client.Object) bool {...}` as a function. You can take inspiration from here: https://github.com/Kong/kubernetes-ingress-controller/blob/fc162862696c8eb4c400a73336abc0017af78f23/internal/controllers/gateway/gateway_controller.go#L91

mlavacca (changes requested):
Thanks for your PR @Rahul-D78!

controllers/gateway_controller.go
```diff
@@ -36,18 +39,35 @@ const gatewayServiceLabel = "konghq.com/owned-by-gateway"
...
+func (r *GatewayReconciler) gatewayHasMatchingGatewayClass(obj client.Object) bool {
+	gateway, ok := obj.(*gatewayv1beta1.Gateway)
+	if !ok {
+		r.Log.Error(fmt.Errorf("unexpected object type in gateway watch predicates"), "expected", "*gatewayv1beta1.Gateway", "found", reflect.TypeOf(obj))
+		return false
+	}
+	gatewayClass := &gatewayv1beta1.GatewayClass{}
+	if err := r.Client.Get(context.Background(), client.ObjectKey{Name: string(gateway.Spec.GatewayClassName)}, gatewayClass); err != nil {
+		r.Log.Error(err, "could not retrieve gatewayclass", "gatewayclass", gateway.Spec.GatewayClassName)
+		return false
```
shaneutt, line 66: Strictly speaking we don't know anything about the error we received at this point. It could be that the object was missing (e.g. `404`) or if the cluster is having a bad day it could actually be a server error (e.g. `5XX`). As such we should try and explicitly check if the error is a `NotFound`:

```suggestion
		if errors.IsNotFound(err) {
				r.Log.Error(err, "gatewayclass not found", "gatewayclass", gateway.Spec.GatewayClassName)
				return false
		}
		return false
```

However if the error is something other than `NotFound` we're in the territory of temporarily failures in the Kubernetes API and such, so returning false for those could result in missing an enqueuing of a valid object while the API server is otherwise having issues, and thus reconciliation would ultimately go stale for that object until it gets another update.

Since the Gateway controller will also do this `GatewayClass` check this predicate is just an optimization for busy clusters, so it's safe and preferable to just enqueue the object anyway when we're not really sure what's going on:

```suggestion
		if errors.IsNotFound(err) {
				r.Log.Error(err, "gatewayclass not found", "gatewayclass", gateway.Spec.GatewayClassName)
				return false
		}
		return true # some other problem retrieving the object, enqueue anyway to avoid dropping
```

As the comment above indicates, I recommend that we return true as a failsafe to avoid "losing track" of objects during API failure scenarios.

shaneutt (changes requested):
Looks great! Thanks for adding this.

I have one comment about optimizing the predicate function, and then :+1:

controllers/gateway_controller.go
Rahul-D78: thanks for the detailed explanation @shaneutt

shaneutt (approved):
LGTM, thank you!

The original issue had requested a unit test for this, but I realized we don't have any prior art. I'm going to create a follow-up PR to add a unit test for this, and then from there we'll have an example we can use for future tests.

Rahul-D78:
Thanks a lot @shaneutt and @mlavacca :+1:

PR #23, opened 2022-11-14
Add TC_INGRESS UDP redirect program

astoycos:
Add a new program which uses the `bpf_redirect_neigh` helper to redirect traffic from a host's main NIC to a container interface. To get this working make sure ip_forwarding is enabled on your kernel, and ensure there are no net-filter rules which could drop the incoming traffic. 

TODO 
- https://github.com/Kong/blixt/issues/24
- https://github.com/Kong/blixt/issues/25
- Update control-plane to use the `Backends` map and actually program the ebpf-program: https://github.com/Kong/blixt/issues/11


shaneutt (dismissed):
Looking good! I added issues for the TODO items so we can follow up. This gets us a step closer to the Rust re-write, I'm going to merge this as is so we can continue hacking on it to plug it into the UDPRoute controller. :+1:

shaneutt (approved):

PR #26, opened 2022-11-18
feat: add grpc api for dataplane

shaneutt:
partially resolves https://github.com/Kong/blixt/issues/11, but now the `UDPRoute` controller will live with the rest of the control-plane.

This also includes a new xtask for manually
inserting and removing routing data from the
dataplane, e.g.:

```console
cargo xtask grpc-client --vip-ip 172.17.0.2 --port 9875 --daddr 10.1.0.25 --dport 9875 --ifindex 7
```

NOTE: This change was made in lieu of adding a
pull mechanism for dataplane configuration (which
is actually what I added before in a previous
iteration). The reason for this choice was purely
to make development easy in the coming weeks,
allowing a quick and easy way to insert routing
data into the dataplane for debugging and testing.
I would not be surprised if later once things are
futher along we went the other way again.


dataplane/ebpf/src/main.rs
```diff
@@ -45,28 +45,26 @@ unsafe fn ptr_at<T>(ctx: &TcContext, offset: usize) -> Result<*mut T, i64> {
     Ok((start + offset) as *mut T)
 }
 
-#[map(name = "BACKENDS")] 
+#[map(name = "BACKENDS")]
```
astoycos, line 48: Eventually we'll probably want this pinned with 

`    HashMap::<BackendKey, Backend>::pinned(128, 0);`

astoycos (approved):
Nice tool to speed up manual control plane debugging! Thanks for doing this!

PR #31, opened 2022-11-29
fix: add protoc to rust build

shaneutt:
Adds `protoc` to the build phase for building the dataplane image as needed for the new Rust based dataplane.


astoycos (approved):
/lgtm thanks!

PR #36, opened 2022-12-04
UDPRoute Controller Scaffolding

rushi47:
Problem Statement 
---
We need to add a controller to handle UDPRoutes. We'll start by adding the
basic scaffolding before we add the dataplane connection. This only includes
the basic scaffolding, not the data-plane configuration push yet that will be
handled in a follow-up.

TODO
---
 - [x] UDPRoute` controller added with `operator-sdk`
 - [x] Logic added to verify `UDPRoute` is attached to a `Gateway` supported by
`GatewayClass` and print this out to logs

Resolves : https://github.com/Kong/blixt/issues/30


controllers/udproute_controller.go
```diff
@@ -0,0 +1,99 @@
...
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+// UDPRouteReconciler reconciles a UDPRoute object
+type UDPRouteReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+// Verify respective gateway has atleast one listener matching to what is specified in UDPRoute parentRefs.
```
shaneutt, line 24:
```suggestion
// Verify respective gateway has at least one listener matching to what is specified in UDPRoute parentRefs.
```

controllers/udproute_controller.go
```diff
@@ -0,0 +1,99 @@
...
+			return ctrl.Result{}, nil
+		}
+		log.Info("Error retrieving udp route", "Err : ", err)
+		return ctrl.Result{}, err
+	}
+
+	//Use the retreive objects its parent ref to look for the gateway.
+	for _, parentRef := range udproute.Spec.ParentRefs {
+
+		//Build Gateway object to retreive
```
shaneutt, line 56:
```suggestion
		//Build Gateway object to retrieve
```

controllers/udproute_controller.go
```diff
@@ -0,0 +1,99 @@
...
+		//Get Gateway for UDP Route
+		if err := r.Get(ctx, types.NamespacedName{Name: string(parentRef.Name), Namespace: ns}, gw); err != nil {
+			log.Info("Unable to get parent ref gateway for", "udpRoute", parentRef.Name, "Namespace", ns, "err", err)
+			return ctrl.Result{}, err
+		}
+
+		//Check if referred gateway has the atleast one listner with properties defined from UDPRoute parentref.
+		if err := r.verifyListener(ctx, gw, parentRef); err != nil {
+			log.Info("No matching listener found for referred gateway", "GatewayName", parentRef.Name, "GatewayPort", parentRef.Port)
+			return ctrl.Result{}, err
```
shaneutt, line 73: So by default, errors will trigger a re-queue. Since this condition (where we don't have a valid listener to attach to) is not a resolvable condition without end-user interference it would be better to stop and update the status of the `UDPRoute`. That would however be better to resolve as a separate smaller PR, so I've created a follow up for this: https://github.com/Kong/blixt/issues/39

controllers/udproute_controller.go
```diff
@@ -0,0 +1,99 @@
...
+			log.Info("Unable to get Gatewayclass", "Gateway", gw.Name, "Udp Route", parentRef.Name, "err", err)
+			return ctrl.Result{}, err
+		}
+
+		if gwc.Spec.ControllerName != vars.GatewayClassControllerName {
+			return ctrl.Result{}, nil
+		}
+
+		log.Info("UDP Route appeared referring to Gateway", "UDPRoute", parentRef.Name, "Gateway ", gw.Name, "GatewayClass, Controller Name", gwc.Spec.ControllerName, "GatewayClass Name", gwc.ObjectMeta.Name)
+	}
```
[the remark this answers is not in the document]
shaneutt: Oh, and if you agree with doing it as a follow-up, please create a follow-up issue, something like "UDPRoute multiple Gateway support" which we link into a `// TODO` comment here.

controllers/udproute_controller.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"github.com/kong/blixt/pkg/vars"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	gatewayv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
```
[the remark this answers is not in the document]
rushi47: gonna find this one looks to be useful

controllers/udproute_controller.go
rushi47: Will need help on this one, resolving here and will start discussion on #39  thread.

controllers/udproute_controller.go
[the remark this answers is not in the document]
rushi47: Done added check and refactored code. 
Also created issue and referred it in code : https://github.com/Kong/blixt/issues/40

shaneutt (dismissed):
Thanks @rushi47!

While there's a couple of things we could still tweak here, there's a few other issues waiting on this one and this controller currently isn't actually fully implemented so this seems like a fine jumping off point and we can start on some follow-up iterations. :+1: 

Appreciate the contribution, good work! :vulcan_salute:

PR #37, opened 2022-12-04
Command substitution

rushi47:
- Cleanup, as there could be case that default shell is not bash.

Signed-off-by: Rushikesh Butley <www.rushi47@gmail.com>


rushi47:
@shaneutt, grabbing your attention here too.

Issue #45, opened 2022-12-12
Basic Egress


shaneutt:
## Problem Statement

[Previously](https://github.com/kong/blixt/issues/11) we added ingress support using TC, but next (and in order to support [TCPRoute](https://github.com/kong/blixt/issues/16)) we will need to be able to return traffic.

## Additional Notes

Technically newer kernels allow access to `ip_conntrack` from within TC programs, so seemingly it's on the table for us to use that.

## Proposed Implementation

- [ ] implement our own basic connection tracking in an eBPF map with SNAT
- [ ] add a new `tc_egress` TC program responsible for getting traffic back out

## Acceptance Criteria

- [ ] ICMP messages for closed ports and closed sockets can return to the original client via `UDPRoute`

PR #52, opened 2022-12-14
feat: add initial UDPRoute support

### LINKED ISSUE #41: UDPRoute Controller GRPC Integration
### Problem Statement

We previously handled data-plane configuration at a different level, but since then added the scaffolding for a `UDPRoute` controller in the control-plane which would be responsible for pushing `UDPRoute` related configurations to the data-plane via its new GRPC API. The purpose of this task is to hook up all the wires to make `UDPRoute` compile to data-plane `Backends`.

### Acceptance Criteria

- [ ] a `UDPRoute` resource should be configured and route properly through the data-plane 
- [ ] the [demo](https://www.youtube.com/watch?v=id3DYxUfqdo) we previously ran should now be runnable again, but now with TC instead of XDP

shaneutt:
This PR integrates the controlplane with the dataplane so as to add the initial, limited support of the [UDPRoute](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1alpha2.UDPRoute) API.

This ended up getting very large, and including several ancillary things:

 - a UDP testing server (with container image, and kustomize samples)
 - added a `--release` image build of the dataplane, now deployed on `alpine`
 - GRPC clients are now automaticallly generated for the controlplane
 - several bug fixes and minor improvements
 - an integration testing suite, ci workflows for the test suite, and an initial integration test for `UDPRoute`

The last one being the most notable: as starting from this PR we will now have meaningful CI tests which can tell us if we're breaking functionality, as this test will deploy the controlplane and dataplane, and then use `UDPRoute` to get traffic to our new UDP testing server in a backend `Pod`.

But there's still a lot more to do: since this PR ended up getting so large there are several follow-ups marked as TODOs each with a link to the relevant Github issue.

Resolves #41


tools/udp-test-server/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+COPY Cargo.toml Cargo.toml
+
+COPY Cargo.lock Cargo.lock
+
+COPY src/main.rs src/main.rs
+
+RUN RUSTFLAGS=-Ctarget-feature=+crt-static cargo build --release --target=x86_64-unknown-linux-musl
+
+FROM gcr.io/distroless/static
```
astoycos, line 19: small nit :) but any reason not to just use `scratch` here?

dataplane/api-server/src/netutils.rs
```diff
@@ -0,0 +1,57 @@
...
+    let ifindex = unsafe { libc_if_nametoindex(ifname_c.into_raw()) };
+    Ok(ifindex)
+}
+
+/// Given an IPv4 address will return the local system's network interface
+/// which is responsible for routing that address. Not portable: only works on
+/// Linux systems with iproute2 installed.
+///
+/// TODO: replace this https://github.com/Kong/blixt/issues/49
+pub fn if_name_for_routing_ip(ip_addr: Ipv4Addr) -> Result<String, Error> {
```
astoycos, line 21: In the future we may be able to interact with netlink similar to what's done here https://github.com/redhat-et/emerita/blob/ade6ca9c1215704da6ce033e0b4e8a5f39dbd908/emerita/src/netlink.rs#L46 just using rtnetlink instead

tools/udp-test-server/Dockerfile
shaneutt: No specific reason not to.

pkg/vars/vars.go
```diff
@@ -1,5 +1,41 @@
...
-// responsible for relevant resources.
-const GatewayClassControllerName = "konghq.com/blixt"
+// -----------------------------------------------------------------------------
+// ControlPlane Constants
+// -----------------------------------------------------------------------------
+
+const (
+	// GatewayClassControllerName is the unique identifier indicating controller
+	// responsible for relevant resources.
+	GatewayClassControllerName = "konghq.com/blixt"
```
astoycos, line 10: This should probably stay vendor agnostic barring the move to sigs right?

pkg/vars/vars.go
```diff
@@ -1,5 +1,41 @@
...
+
+	// DefaultNamespace is the namespace used for the controlplane and
+	// dataplane components (by default).
+	DefaultNamespace = "blixt-system"
+
+	// DefaultDataPlaneAPIPort is the network port that will be used to
+	// communicate with the DataPlane API (by default).
+	DefaultDataPlaneAPIPort = 9874
+
+	// DefaultDataPlaneAppLabel indicates the label value that can be used
```
astoycos, line 34: /the label value/ -> /the label key/ ? 

Or am I miss-understanding

internal/dataplane/client/client.go
```diff
@@ -0,0 +1,65 @@
...
+	endpoints, err := GetDataPlaneEndpointsFromDefaultPods(ctx, c)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(endpoints) < 1 {
+		return nil, fmt.Errorf("no endpoints could be found for the dataplane API")
+	}
+
+	if len(endpoints) > 1 {
```
astoycos, line 26: oooh this will be fun :) Any issues to track this?

controllers/utils.go
```diff
@@ -0,0 +1,28 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// DataPlaneFinalizer is the finalizer which indicates that an object needs to
+	// have its configuration removed from the dataplane before it can be deleted.
+	DataPlaneFinalizer = "blixt/dataplane-configuration"
+)
+
+func isDataPlaneFinalizerSet(obj client.Object) bool {
```
astoycos, line 15: `controllerutil.ContainsFinalizer`  from "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 Has this already done for yah :)

config/dataplane/dataplane.yaml
```diff
@@ -4,22 +4,27 @@ metadata:
...
     spec:
-      serviceAccountName: udproute-manager
       hostNetwork: true
       containers:
       - name: dataplane
         image: ghcr.io/kong/blixt-dataplane:latest
         securityContext:
           privileged: true
-        args: ["eth0"]
+        args: ["-i", "eth0"]
```
astoycos, line 26: Should this be configurable?

test/integration/suite_test.go
```diff
@@ -9,36 +9,185 @@ import (
 	"os"
 	"testing"
 
+	"github.com/kong/kubernetes-testing-framework/pkg/clusters"
```
astoycos, line 12: I'm pretty unfamiliar with this tool but it sure does make testing easier :)

test/integration/udproute_test.go
```diff
@@ -0,0 +1,77 @@
...
+	require.Len(t, pods.Items, 1)
+	udpServerPod := pods.Items[0]
+	req := env.Cluster().Client().CoreV1().Pods(corev1.NamespaceDefault).GetLogs(udpServerPod.Name, &corev1.PodLogOptions{})
+	logs, err := req.Stream(ctx)
+	require.NoError(t, err)
+	defer logs.Close()
+	output := new(bytes.Buffer)
+	_, err = io.Copy(output, logs)
+	require.NoError(t, err)
+	require.Contains(t, output.String(), message)
```
astoycos, line 76: Eventually It'd be cool to verify the srcIP here too (I think the test server dumps that as well? )

astoycos:
Overall /lgtm I just really had a bunch of NITs and questions :)

astoycos (approved):

astoycos:
(one more small nit...can you make the integration tests required?)

dataplane/api-server/src/netutils.rs
shaneutt: added a note about this to https://github.com/Kong/blixt/issues/49

pkg/vars/vars.go
shaneutt: This is what it was already set to it just moved. I think we can just make one big sweeping PR to change the naming once it fully transitions.

pkg/vars/vars.go
shaneutt: No, it is indeed the label value.

internal/dataplane/client/client.go
shaneutt: Yes, https://github.com/Kong/blixt/issues/10 tracks this.

config/dataplane/dataplane.yaml
shaneutt: Eventually I expect we may remove this argument entirely and allow the loader to determine the interfaces it needs to deploy ebpf programs based on some dynamic criteria. For the moment, this just fixes something that was already broken, I think once we get to #45 we'll start having a better idea about how this is gonna look going forward.

test/integration/suite_test.go
shaneutt: Yeah that's why it exists. I think there's a chance we'd be open to giving it to upstream, but I'm pretty sure the maintenance of trying to be "the testing framework for kubernetes" would be intense.

test/integration/udproute_test.go
shaneutt: Right now developers are free to provide their own kind clusters with different network configurations than what we have, so it can be a little complicated to know what the source IP is going to be ahead of time. I do agree it would be good to verify all the datapoints we can eventually, but at least for the moment since this PR already got so large I would like to hold back. I created https://github.com/Kong/blixt/issues/53 as a follow-up to add this later.

controllers/utils.go
shaneutt: d989894183724bbd39b163dcb68984dd77d4111c adds this :+1:

PR #54, opened 2022-12-15
chore: various cleanup

shaneutt:
- chore: rename controller-manager to controlplane
- chore: cleanup the README.md
- chore: improve .dockerignore files throughout
- fix: udp test server ignore stream data
- fix: reduce spam from healthcheck peers
- chore: improve caching for rust image builds
- feat: add makefile directive to build dev cluster

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,8 @@ After these goals are achieved, further goals will be decided.
 ## Usage
 
 > **Note**: Currently usage is only possible on [Kubernetes In Docker
-> (KIND)][kind] clusters.
+> (KIND)][kind] clusters. You can generate a new development cluster for
+> testing with `make build.cluster`.
 
 Deploy [Gateway API][gwapi] [CRDs][crds]:
 
--- a/README.md
+++ b/README.md
@@ -55,10 +55,13 @@ After these goals are achieved, further goals will be decided.
 
 ## Usage
 
+> **Note**: Currently usage is only possible on [Kubernetes In Docker
+> (KIND)][kind] clusters.
+
 Deploy [Gateway API][gwapi] [CRDs][crds]:
 
 ```console
-kubectl kustomize https://github.com/kubernetes-sigs/gateway-api/config/crd/experimental | kubectl apply -f -
+kubectl kustomize https://github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.5.1 | kubectl apply -f -
 ```
 
 Deploy:
@@ -67,6 +70,20 @@ Deploy:
 kubectl kustomize config/default | kubectl apply -f -
 ```
 
+At this point you should see the `controlplane` and `dataplane` pods running
+in the `blixt-system` namespace:
+
+```console
+$ kubectl -n blixt-system get pods
+NAME                                 READY   STATUS    RESTARTS   AGE
+blixt-controlplane-cdccc685b-9dxj2   2/2     Running   0          83s
+blixt-dataplane-brsl9                1/1     Running   0          83s
+```
+
+Check the `config/samples` directory for `Gateway` and `*Route` examples you
+can now deploy.
+
+[kind]:https://github.com/kubernetes-sigs/kind
 [gwapi]:https://github.com/kubernetes-sigs/gateway-api
 [crds]:https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/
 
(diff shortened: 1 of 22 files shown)


README.md
```diff
@@ -67,6 +71,20 @@ Deploy:
 kubectl kustomize config/default | kubectl apply -f -
 ```
 
+At this point you should see the `controlplane` and `dataplane` `Pods` running
```
astoycos, line 74:
```suggestion
At this point you should see the `controlplane` and `dataplane` pods running
```

README.md
```diff
@@ -67,6 +71,20 @@ Deploy:
...
+in the `blixt-system` namespace:
+
+```console
+$ kubectl -n blixt-system get pods
+NAME                                 READY   STATUS    RESTARTS   AGE
+blixt-controlplane-cdccc685b-9dxj2   2/2     Running   0          83s
+blixt-dataplane-brsl9                1/1     Running   0          83s
+```
+
+Check the `config/samples` directory for `Gateway` and `*Route` examples you
```
astoycos, line 84:
```suggestion
Check the `config/samples` directory for `Gateway` and `Route` examples you
```

dataplane/Dockerfile
```diff
@@ -7,14 +7,20 @@ RUN rustup default stable
 RUN rustup install nightly
 RUN rustup component add rust-src --toolchain nightly
 RUN rustup target add x86_64-unknown-linux-musl
-RUN cargo install bpf-linker
+RUN --mount=type=cache,target=/root/.cargo/registry \
```
astoycos, line 10: oooh will this actually cache the install of bpd-linker? That's nice :)

dataplane/Dockerfile
```diff
@@ -7,14 +7,20 @@ RUN rustup default stable
...
-RUN --mount=type=cache,target=/workspace/target/ cargo xtask build-ebpf --release
-RUN --mount=type=cache,target=/workspace/target/ RUSTFLAGS=-Ctarget-feature=+crt-static cargo build --release --target=x86_64-unknown-linux-musl
-RUN --mount=type=cache,target=/workspace/target/ cp /workspace/target/x86_64-unknown-linux-musl/release/loader /workspace/dataplane
+RUN --mount=type=cache,target=/workspace/target/ \
+    --mount=type=cache,target=/root/.cargo/registry \
+    cargo xtask build-ebpf --release
+RUN --mount=type=cache,target=/workspace/target/ \
+    --mount=type=cache,target=/root/.cargo/registry \
+    RUSTFLAGS=-Ctarget-feature=+crt-static cargo build --release --target=x86_64-unknown-linux-musl
+RUN --mount=type=cache,target=/workspace/target/ \
```
astoycos, line 22: Could you just copy the binary on line 32 from the builder stage directly rather than doing the extra copy here?

astoycos (approved):
Some Smaller NITS otherwise lgtm

mlavacca (approved):

README.md
shaneutt: Well, in this case I'm trying to actually indicate that we're talking about Kubernetes `Pods` which is why it's highlighted this way, and CamelCased.

README.md
shaneutt: This was intentional, in Gateway API we often refer to the various route types as `*Routes` or `xRoutes`.

dataplane/Dockerfile
shaneutt: Yes, which saves a LOT of time

dataplane/Dockerfile
shaneutt: Unfortunately no because of the use of the buildkit cache. The final binary is actually not present in the filesystem layers of the image, UNLESS I copy it out of the build cache. I could probably do some other tricks to make this work, like caching differently but I think this is the simplest solution.

dataplane/Makefile
```diff
@@ -20,3 +21,8 @@ build.release:
 .PHONY: build.image
 build.image:
 	DOCKER_BUILDKIT=1 docker build -t $(IMAGE):$(TAG) ./
+
+.PHONY: load.image
+load.image: build.image
+	kind load docker-image $(IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
+		kubectl -n blixt-system rollout restart daemonset blixt-dataplane
```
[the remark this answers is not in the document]
shaneutt: I'm indenting here because this command is chained to the success of the previous command, it is a one-liner

README.md
shaneutt: updated in a00556c430e45bd4b84efd6f05a0e4fb3ea99c32

Issue #15, opened 2022-11-08
TCP Support (data-plane)


shaneutt:
## Problem Statement

Parts of TCP support already exist, but the dataplane doesn't have full support yet. The purpose of this task is to get a basic level of TCP support added to the dataplane.

## Acceptance Criteria

- [x] `TCP` implemented to the extent that HTTP traffic is working
- [x] add integration tests for `TCPRoute`

PR #55, opened 2022-12-20
feat: add TCPRoute controller

### LINKED ISSUE #16: TCPRoute Support (control-plane)
## Problem Statement

After we enable TCP support in the data-plane we should expose that support via the Gateway API resource [TCPRoute]().

## Acceptance Criteria

- [x] a `TCPRoute` controller is added which can compile `BACKEND` resources and push them to the `BACKENDS` map

shaneutt:
This primarily adds the `TCPRoute` controller and samples. Note however that the dataplane does not have complete TCP support yet (though some scaffolding for it is added as of this PR) so the dataplane configuration produced is valid but the samples wont work yet. This scaffolding however will help to support #45 and #15 which are responsible for completing this.

Resolves #16

This also includes some refactoring and cleanup, notably it refactors the dataplane code into modules.


controllers/tcproute_controller.go
```diff
@@ -0,0 +1,231 @@
...
+		gw := new(gatewayv1beta1.Gateway)
+
+		ns := tcproute.Namespace
+		if parentRef.Namespace != nil {
+			ns = string(*parentRef.Namespace)
+		}
+
+		//Get Gateway for TCP Route
+		if err := r.Get(ctx, types.NamespacedName{Name: string(parentRef.Name), Namespace: ns}, gw); err != nil {
+			log.Info("Unable to get parent ref gateway for", "tcpRoute", parentRef.Name, "Namespace", ns, "err", err)
```
mlavacca, line 125: I think we should check if the error is `NotFound`; in that case, keeping iterating to find a suitable gateway is fine. If we encounter another error, maybe we should immediately return it as an unexpected error occurred. WDYT?

controllers/tcproute_controller.go
```diff
@@ -0,0 +1,231 @@
...
+
+		supportedGateways = append(supportedGateways, *gw)
+	}
+
+	//only used for logging
+	if len(supportedGateways) < 1 {
+		return false, nil, parentRefErr
+	}
+
+	refferedGateway := &supportedGateways[0]
```
mlavacca, line 160:
```suggestion
	referredGateway := &supportedGateways[0]
```

controllers/tcproute_controller.go
```diff
@@ -0,0 +1,231 @@
...
+		}
+		r.log.Info("Error retrieving tcp route", "Err : ", err)
+		return ctrl.Result{}, err
+	}
+
+	isManaged, gateway, err := r.isTCPRouteManaged(ctx, *tcproute)
+	if !isManaged {
+		// TODO: enable orphan checking https://github.com/Kong/blixt/issues/47
+		return ctrl.Result{}, err
+	}
```
mlavacca, line 71:
```suggestion
	isManaged, gateway, err := r.isTCPRouteManaged(ctx, *tcproute)
	if err != nil {
		return ctrl.Result{}, err
	}
	if !isManaged {
		// TODO: enable orphan checking https://github.com/Kong/blixt/issues/47
		return ctrl.Result{}, nil
	}
```

mlavacca: For the sake of clarity, I think we should explicitly check the error here, because, as I get from the code, isManaged can be false with err == nil (no suitable gateway has been found) or with error != nil (an error occurred when getting the gateways, and no suitable gateway has been found).

controllers/tcproute_controller.go
```diff
@@ -0,0 +1,231 @@
...
+
+	//only used for logging
+	if len(supportedGateways) < 1 {
+		return false, nil, parentRefErr
+	}
+
+	refferedGateway := &supportedGateways[0]
+	log.Info("TCP Route appeared referring to Gateway", "Gateway ", refferedGateway.Name, "GatewayClass Name", refferedGateway.Spec.GatewayClassName)
+
+	return true, refferedGateway, nil
```
mlavacca, line 163: Why do we want to return only the first Gateway of the list? Shouldn't we take into account all the matching Gateways?

controllers/tcproute_controller.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// delete the target from the dataplane
+	confirmation, err := dataplaneClient.Delete(context.Background(), targets.Vip)
+	if err != nil {
+		return err
+	}
+
+	r.log.Info(fmt.Sprintf("successful data-plane DELETE, confirmation: %s", confirmation.String()))
+
+	oldFinalizers := tcproute.GetFinalizers()
```
mlavacca, line 221: WDYT about moving the finalizer removal out of this function? The function is named `ensureTCPRouteDeletedInDataPlane`, and removing also the dataplane finalizers looks a bit out of context.

tools/udp-test-server/Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN --mount=type=cache,target=/workspace/target/ \
 RUN --mount=type=cache,target=/workspace/target/ \
     cp /workspace/target/x86_64-unknown-linux-musl/release/udp-test-server /workspace/udp-test-server
 
-FROM gcr.io/distroless/static
+FROM alpine
```
mlavacca, line 21: Why are we moving this one from distroless to alpine?

mlavacca (commented):
I left some comments on the controlplane part of the PR, I'll let @astoycos chime in and review the rust/eBPF side of things.

controllers/tcproute_controller.go
shaneutt: I think it's entirely relevant, the entire purpose of the finalizer is to hold up the object until we've confirmed removal from the dataplane.

controllers/tcproute_controller.go
shaneutt: (keep in mind that a lot of this controller code is copy/paste from the `UDPRoute` controller)

Yep this is a good suggestion, but it will enable a race condition that can cause a stale route so I've fixed that by adding a `Gateway`->`*Route` map func to both the `TCPRoute` and `UDPRoute` controllers to help compensate:

b1be1b75316230e4088b52fea6e5fe44082e5aaf
a6308da1f5e7e0de4cb95d6c6d6a569a76a8740d

And I've refactored both this method, and the method it was copy/pasted from in the `UDPRoute` controller:

f00e818f61908df9b9fa28d875ab29aa6a121c2f

And they no longer bother collecting errors anymore, any error that is not a 404 is considered a total failure to reconcile.

controllers/tcproute_controller.go
shaneutt: Another copy/paste from the old controller, now updated in both:

ffe1583221d679908b9ae7707bd7da7bbd6501db

controllers/tcproute_controller.go
shaneutt: In the code's current form, this wont make any technical difference, but I'm with you that the clarity could be better for the future. As this was copy/paste from the udproute controller, updated both in:

fa3af62270d21f1d6186ef4cac58e0b8ec13f273

controllers/tcproute_controller.go
shaneutt: Yes, but we're pacing ourselves to add support for multiple `Gateways` later as we don't need it right this moment for testing, and it provides an opportunity for a community member to jump in and work on something that's not huge. @rushi47 created https://github.com/Kong/blixt/issues/40 as a follow up for this, but it looks like we forgot to link that here. I've added the TODO with the link to both controllers in:

497953abbe09cd0052cf78086422b15092235bcc

controllers/tcproute_controller.go
shaneutt: I think it's entirely relevant, the entire purpose of the finalizer is to hold up the object until we've confirmed removal from the dataplane. Keep in mind too that now that we're moving this project to Kubernetes SIGs it is intended to become a reference implementation for Gateway API, and as such it's going to undergo a LOT of refactoring and reviews from Gateway API community members before we start advertising it as a reference implementation (in fact, the intention right now is that before we consider it "ready" as a reference implementation we will do a full PR review of the code with the org).

tools/udp-test-server/Dockerfile
shaneutt: The commit for this included an explanation:

d783a56dc72de73b5b04b7099f3a09df175855be

```
This enables the ability to add tools, such as tcpdump
and do some debugging.
```

shaneutt:
added `do-not-merge` as I intend to do some commit squashing in rebase before we merge.

mlavacca (approved):

mlavacca (dismissed):

mlavacca (approved):

dataplane/ebpf/src/main.rs
```diff
@@ -98,77 +60,36 @@ fn try_tc_ingress(ctx: TcContext) -> Result<i32, i64> {
...
-        daddr_dot_dec[2],
-        daddr_dot_dec[3],
-        u16::from_be(unsafe { (*udp_hdr).dest })
-    );
-
-    // Update destination IP
-    unsafe {
-        (*ip_hdr).daddr = backend.daddr.to_be();
+    match protocol {
+        IPPROTO_TCP => handle_tcp_ingress(ctx),
```
astoycos, line 64: I think it's still smarter to handle any errors we may encounter in our `handle` functions here rather than internally

astoycos: (i.e like we still do for `tc_egress`)

astoycos: Oh nm we do :) in `tc_ingress` my bad

Makefile
```diff
@@ -268,3 +268,8 @@ KIND_CLUSTER ?= blixt-dev
 .PHONY: build.cluster
 build.cluster: $(KTF) # builds a KIND cluster which can be used for testing and development
 	PATH="$(LOCALBIN):${PATH}" $(KTF) env create --name $(KIND_CLUSTER) --addon metallb
+
+.PHONY: load.image
+load.image: build.image
+	kind load docker-image $(IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
+		kubectl -n blixt-system rollout restart deployment blixt-controlplane
```
astoycos, line 275:
```suggestion
	kubectl -n blixt-system rollout restart deployment blixt-controlplane
```

shaneutt: This is a chained command with the previous line, thus why we're indenting as a visual cue

shaneutt:
Re-based, ready to merge :+1:

PR #57, opened 2022-12-22
chore: experimental proto3 flag

mlavacca:
With Ubuntu 22.04, the latest version (installed via apt-get) of protobuf is `3.12.4`. With such a version, the dataplane build requires the flag `--experimental_allow_proto3_optional` to be successful. This PR introduces such a flag in both the api-server and xtask build targets.


shaneutt (approved):
I checked this out locally and it appears to work fine with my `protoc` version `3.21.12` and I made some changes and tested that generation was working properly, and it seems to be.

Thanks, LGTM!

PR #60, opened 2023-01-09
Add basic Conntrack for UDP

### LINKED ISSUE #45: Basic Egress

astoycos:
Works to fix #45 

Specifically this patch adds another program to the egress interface which catches ICMP dest unreachable packets and re-transmits them back to the original client.

Still working on the integration testing bits to prove that this works in K8s


dataplane/ebpf/src/egress/icmp.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    // We only care about redirecting port unreachable messages currently so a 
+    // UDP client can tell when the server is shutdown
+    if unsafe { (*icmp_hdr).type_ } != ICMP_PROTO_TYPE_UNREACH { 
+        return Ok(TC_ACT_PIPE);
+    }   
+
+    let dest_addr = unsafe { (*ip_hdr).daddr };
+    
+    let new_src = unsafe { CONNTRACK.get(&dest_addr) }.ok_or(TC_ACT_PIPE)?;
```
shaneutt, line 39: I almost wonder if it would be worth naming our "conntrack" something more unique to us, to avoid any confusion that this is our own implementation. What do you think about `BLIXT_CONNTRACK`?

dataplane/ebpf/src/egress/icmp.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    let daddr_dot_dec = ip_from_int(unsafe { (*ip_hdr).daddr });
+    info!(
+        &ctx,
+        "Received a ICMP Unreachable packet destined for svc ip: {}.{}.{}.{}",
+        daddr_dot_dec[0],
+        daddr_dot_dec[1],
+        daddr_dot_dec[2],
+        daddr_dot_dec[3],
+    );
```
shaneutt, line 49: It occurs to me that I added this for convenience (and boy is it an improvement) but we can probably add one more wrapper on top of this to clean it up further so we don't have to enumerate the indices in every call. Doesn't need to be done in this PR though.

shaneutt (commented):
Cool, I know it's WIP but looks like we've got a start on something that can work and we could demo soon :+1: one step along the way to the eventual TCP support :fist_right::fist_left:

dataplane/ebpf/src/egress/icmp.rs
astoycos: I'll do it :)

dataplane/ebpf/src/egress/icmp.rs
astoycos: Totally agreed let me make an issue for it.

config/samples/udproute-noreach/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
...
+  metadata:
+    name: blixt-udproute-sample
+  spec:
+    template:
+      spec:
+        containers:
+        - name: server
+          command:
+            - ./udp-test-server
+            - --dry-run
```
shaneutt, line 17: Should we add a comment on this as a breadcrumb to the user as the entire point of this patch is to apply this?

Maybe something like:

```suggestion
            # --dry-run disables UDP listeners in order to test failures to send data, and trigger ICMP port failure responses from the kernel
            - --dry-run
```

dataplane/ebpf/src/bindings.rs
```diff
@@ -1,4 +1,4 @@
-/* automatically generated by rust-bindgen 0.63.0 */
+/* automatically generated by rust-bindgen 0.61.0 */
```
shaneutt, line 1: Seems like we should continue using the latest release, even if it doesn't seem to make a difference? I don't think it should change anything below but might as well right?

dataplane/ebpf/src/egress/icmp.rs
```diff
@@ -0,0 +1,93 @@
...
+            icmp_inner_ip_hdr as *mut u32,
+            mem::size_of::<iphdr>() as u32,
+            0
+        )
+    } as u64;
+    unsafe { (*icmp_inner_ip_hdr).check = csum_fold_helper(full_cksum) };
+
+    // remove conntrack entry after icmp port unreachable message is sent back
+    // to the client.
+    // TODO(astoycos) we should handle these failure modes better
```
shaneutt, line 89: We should do something about this TODO, whether that be resolving it now by improving the failure mode handling, OR by creating a follow-up issue and linking it here in the comment so that the TODO does not go out-of-band.

dataplane/ebpf/src/egress/icmp.rs
```diff
@@ -0,0 +1,93 @@
...
+    unsafe { BLIXT_CONNTRACK.remove(&dest_addr)? };
+
+    return Ok(TC_ACT_PIPE);
+}
```
shaneutt, line 93: As it pertains to https://github.com/aya-rs/aya/discussions/505 our ingress/egress code is becoming quite complex, and uses a fair amount of `info()` calls. Given https://github.com/aya-rs/aya/discussions/505#discussioncomment-4806350 perhaps we should take some time and refactor more of these into a several functions?

I'm fine with making that refactoring a follow-up issue, actually that probably would give others a fairly straightforward thing to work on if they're hankering to work on the dataplane code.

dataplane/ebpf/src/main.rs
```diff
@@ -78,12 +85,28 @@ pub fn tc_egress(ctx: TcContext) -> i32 {
         Err(_) => TC_ACT_SHOT,
     };
 
+    // TODO(astoycos) better Error reporting framework
```
shaneutt, line 88: Let's create a follow-up issue for these TODOs and link them here, yeah?

test/integration/udproute_test.go
```diff
@@ -75,3 +90,62 @@ func TestUDPRouteBasics(t *testing.T) {
...
+	})
+
+	// TODO (astoycos) this currently won't work but it will with updated control-plane logic
+	// add it back so we don't have to maintain a whole new kustomize config for a one line change.
+	// // Update Server to ensure it's in dry run mode then wait for it to be ready
+	// server, err := env.Cluster().Client().AppsV1().Deployments(corev1.NamespaceDefault).Get(ctx, udprouteSampleName, metav1.GetOptions{})
+	// require.NoError(t, err)
+	// server.Spec.Template.Spec.Containers[0].Command = []string{"./udp-test-server", "--dry-run"}
+	// _, err = env.Cluster().Client().AppsV1().Deployments(corev1.NamespaceDefault).Update(ctx, server, metav1.UpdateOptions{})
+	// require.NoError(t, err)
```
shaneutt, line 115: This TODO would appear to warrant a follow-up issue.

shaneutt (changes requested):
Mostly looks great! My comments are mostly minor things and for creating follow-up issues :+1:

.github/workflows/test.yaml
```diff
@@ -67,3 +67,13 @@ jobs:
       env:
         BLIXT_CONTROLPLANE_IMAGE: "ghcr.io/kong/blixt-controlplane:integration-tests"
         BLIXT_DATAPLANE_IMAGE: "ghcr.io/kong/blixt-dataplane:integration-tests"
+        BLIXT_UDP_SERVER_IMAGE: "ghcr.io/kong/blixt-udp-test-server:integration-tests"
+
+    ## Upload diagnostics if integration test step failed.
+    - name: upload diagnostics
```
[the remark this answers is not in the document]
astoycos: NP!

config/samples/udproute-noreach/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
+
```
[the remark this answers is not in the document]
astoycos: I agree, I think I'd rather have this under `/config/test` as it does seem to be more relevant

dataplane/ebpf/src/bindings.rs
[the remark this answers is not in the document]
astoycos: 100% This is just my local version being out of date, let me fix

config/samples/udproute-noreach/kustomization.yaml
astoycos: Yep good Call I will add

test/integration/udproute_test.go
astoycos: This TODO isn't super relevant anymore I like how the Kustomize config turned out with using patches.

dataplane/ebpf/src/egress/icmp.rs
astoycos: https://github.com/Kong/blixt/issues/70

shaneutt (approved):

dataplane/ebpf/src/egress/icmp.rs
astoycos: https://github.com/Kong/blixt/issues/71

PR #61, opened 2023-01-20
fix: pin Rust nightly version to 2023-01-10

synthe102:
The 2023-01-11 nightly rust build [broke aya](https://github.com/aya-rs/aya/issues/490) and subsequently the build of the Blixt Dataplane.
Pinning the rust toolchain version to 2023-01-10 as a temporary fix.

### DIFF
--- a/dataplane/ebpf/rust-toolchain.toml
+++ b/dataplane/ebpf/rust-toolchain.toml
@@ -1,2 +1,3 @@
 [toolchain]
-channel="nightly"
+#TODO: revert once this issue with aya and rust nightly is fixed: https://github.com/aya-rs/aya/issues/490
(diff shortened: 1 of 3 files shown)


dataplane/ebpf/rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel="nightly"
+channel="nightly-2023-01-10"
```
shaneutt, line 2: Let's add a `TODO` comment here with a link to the URL

shaneutt (approved):
One comment about adding the link as a TODO, and then LGTM thank you!

shaneutt (approved):

Issue #71, opened 2023-02-08
Cleanup ip address print helper in the dataplane


astoycos:
Currently to print a human readable ip address in the dataplane we do something like the following 

```
    let daddr_dot_dec = ip_from_int(unsafe { (*ip_hdr).daddr });
    info!(
        &ctx,
        "Received a ICMP Unreachable packet destined for svc ip: {}.{}.{}.{}",
        daddr_dot_dec[0],
        daddr_dot_dec[1],
        daddr_dot_dec[2],
        daddr_dot_dec[3],
    );
```

Let's add a wrapper function on top of this to clean things up even more.

rushi47:
@shaneutt @astoycos can I work on this, as I see its small PR I can test my rust skills to see it in action

shaneutt:
Sounds great @rushi47! Check in with us in here, or in Kubernetes slack if you run into problems!

rushi47:
thanks @shaneutt :)

PR #73, opened 2023-02-25
Helper function to log incoming request

rushi47:
Fixes : #71

### DIFF
--- a/config/samples/tcproute/server.yaml
+++ b/config/samples/tcproute/server.yaml
@@ -16,8 +16,11 @@ spec:
     spec:
       containers:
       - name: server
-        image: nginx
+        image: ghcr.io/shaneutt/malutki
         imagePullPolicy: IfNotPresent
+        env:
+        - name: LISTEN_PORT
+          value: "9080"
         ports:
         - containerPort: 9080
           protocol: TCP


rushi47:
@astoycos  I tried this but this is currently not working.
But this was one approach i was thinking.
Other was adding this method inside :
https://github.com/Kong/blixt/blob/main/dataplane/ebpf/src/bindings.rs#L111 as one of the function.
But i thought it is not the right place and utility is only correct.

dataplane/ebpf/src/utils.rs
```diff
@@ -51,6 +61,20 @@ pub fn ip_from_int(ip: u32) -> [u8; 4] {
     addr
 }
 
+//Return logging string
+pub fn stringify_destip_for_logging(daddr: u32, protocol_code: u8, port: u16) -> String {
```
rushi47, line 65: Apologies but as am newbie :( i am not sure how do I import `String` and `format` in this crate. As root module specified `no_std` i think i can't use this functions and core module seems to only provide `core::str`

astoycos: So unfortunately we'll need to work with raw bytes i,e `[u8]` for any code actually being loaded into the kernel (which includes these util functions) since we can't use any types from the standard lib. 

Now that I dug a bit deeper though... Can you try out this aya-log feature https://github.com/aya-rs/aya/commit/83ec27f06b6859f455f2b2baf985b8fd3fb4adc5? 

Here's an example of it being used -> https://github.com/vadorovsky/aya-examples/blob/main/tc/tc-ebpf/src/main.rs#L40

That might allow us to remove this alltogether

astoycos: cc @shaneutt

Issue #48, opened 2022-12-14
UDPRoute deletion grace period


shaneutt:
## Problem Statement

During #41 we did not add handling for deletion grace period for `UDPRoutes`, instead as a temporary measure just considering any `UDPRoute` deleted if it had any deletion timestamp. The purpose of this task is to add handling for deletion grace periods.

## Acceptance Criteria

- [ ] a deleted `UDPRoute` with a grace period should remain configured in the dataplane until it's grace period ends

PR #80, opened 2023-03-06
Issue 48

rushi47:
Fixes : #48


shaneutt:
Thanks @rushi47!

I think we're slightly off course, but no worries just a clarification: looks like we're trying to reference the `Deployment`'s terminationGracePeriod for pods and whatnot, but all we really need to do is just see if the deletion timestamp is in the past or in the future, having nothing to do with the `Deployment`.

To expand on it: tight now we just check if it _has_ a deletion timestamp, and then we delete. This could be wrong in the case that the timestamp is in the future so we should wait to tear it down until the deletion timestamp has passed. Let me know if that helps?

rushi47:
> Thanks @rushi47!
> 
> 
> 
> I think we're slightly off course, but no worries just a clarification: looks like we're trying to reference the `Deployment`'s terminationGracePeriod for pods and whatnot, but all we really need to do is just see if the deletion timestamp is in the past or in the future, having nothing to do with the `Deployment`.
> 
> 
> 
> To expand on it: tight now we just check if it _has_ a deletion timestamp, and then we delete. This could be wrong in the case that the timestamp is in the future so we should wait to tear it down until the deletion timestamp has passed. Let me know if that helps?

Oh ok got it, thanks for clarification shane. I was thinking like i will get termination period from deployment cause that's the src of truth where we are routing traffic. Then keep this pod alive till the time our pods are there and then terminate it. But i guess like you said this could be simply done with deletion time stamp.

rushi47:
Hey @shaneutt , 
So I was scratching my head over this 😓 ,
And like you suggested i kind of wrote this very raw :
```
	if udproute.DeletionTimestamp != nil {
		r.log.Info("Deleting the Config ...")

		if !(udproute.DeletionTimestamp.After(time.Now())) {
			r.log.Info("Time hasnt passed yet, reconcile..")
			return ctrl.Result{}, nil
		}
		return ctrl.Result{}, r.ensureUDPRouteDeletedInDataPlane(ctx, udproute, gateway)
	}
```
When under deletion it checks if deletion time has passed or not. But I realised, problem with this is, we can’t Reconcile this object expect it to come to this place later as for deletion. If we are here once that means : deletionTimestamp is set. 
Now if the deletiontimestamp is set on resource from my understanding, our client : https://github.com/Kong/blixt/blob/main/controllers/udproute_controller.go#L62
returns object notFound, because of the label deletiontimestamp. 

Maybe to solve this what I think we can do is, to register goroutine or if there is CRD native way to do it maybe we can do that. In this case if we just set graceperiod our config from dataplane will still be deleted though resource will still exist. As there is no way rn, for us to execute some code in dataplane for after specific time.
Maybe we need some thing like this : https://github.com/virtual-kubelet/virtual-kubelet/blob/5a39c167a67dfc636325f2bc07aac96b5f5157da/node/podcontroller.go#L494

Please lmk, if am thinking correctly, apologies for picking your head over this small issue.

shaneutt:
@rushi47 yeah, if I'm understanding correctly what you're getting at your right: part of the problem we need to solve here is to make sure the object doesn't go stale and never become deleted. There's an easy way to do this. When you return `ctrl.Result{}, nil`, it's possible in the `ctrl.Result{}` to use the following fields:

- `Requeue`
- `RequeueAfter`

To ensure that an object intended for deletion continues to get reconciled until it's gone. Historically for similar situations I've done something like this:

```go
	if udproute.DeletionTimestamp != nil {
		if !(udproute.DeletionTimestamp.After(time.Now())) {
			r.log.Info("UDPRoute is set for deletion in the future, waiting to delete", "namespace", req.Namespace, "name", req.Name)
			return ctrl.Result{
				Requeue:        true,
				RequeueAfter: time.Until(gateway.DeletionTimestamp.Time),
			}, nil
		}

		r.log.Info("UDPRoute is being deleted, removing from dataplane", "namespace", req.Namespace, "name", req.Name)
		return ctrl.Result{}, r.ensureUDPRouteDeletedInDataPlane(ctx, udproute, gateway)
	}
```

> **NOTE**: just something to keep in mind, the `ensureUDPRouteDeletedInDataPlane` importantly removes the object's [finalizer](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) as well.

Don't forget that we'll need to add this behavior to both `UDPRoute` _and_ `TCPRoute` (feel free to generalize the functionality if you feel so inclined).

It will be good to add an integration test here to verify the behavior. Ideally the test follows this process:

1. marks an object for deletion several seconds into the future
2. verify that the controller logs the `is set for deletion in the future` line indicating that it's seen the deletion and will wait
3. verify that the controller logs the `removing from dataplane` line indicating final removal
4. for extra thorough-ness use the [gRPC client](https://github.com/Kong/blixt/tree/main/internal/dataplane/client) to communicate with the dataplane and verify the full removal of the configuration data

Let me know if you have any questions, or how we can continue to help!

shaneutt:
@rushi47 if I'm understanding correctly what you're getting at, yes you're right: part of the problem we need to solve here is to make sure the object doesn't go stale and never become deleted (unless some other update happens later). You can often do this using the options in the `ctrl.Result{}`. Check the `godoc` for the following fields:

- `Requeue`
- `RequeueAfter`

In other projects I've done something like this:

```go
	if udproute.DeletionTimestamp != nil {
		if !(udproute.DeletionTimestamp.After(time.Now())) {
			r.log.Info("UDPRoute is set for deletion in the future, waiting to delete", "namespace", req.Namespace, "name", req.Name)
			return ctrl.Result{
				Requeue:        true,
				RequeueAfter: time.Until(gateway.DeletionTimestamp.Time),
			}, nil
		}

		r.log.Info("UDPRoute is being deleted, removing from dataplane", "namespace", req.Namespace, "name", req.Name)
		return ctrl.Result{}, r.ensureUDPRouteDeletedInDataPlane(ctx, udproute, gateway)
	}
```

> **NOTE**: just something to keep in mind, the `ensureUDPRouteDeletedInDataPlane` importantly removes the object's [finalizer](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) as well.

However I'm not sure whether we want to do _exactly_ this in this case: now that we're cleaning up the deletion timestamp handling we're adding a period of time where the configuration is active, despite knowing that we will eventually delete it. We need to make a decision about how we handle _updates_ during this grace period. Personally I think we should continue to handle updates normally, so what we _may_ want to do is set fallback `ctrl.Result{}` options to ensure any result (such as successful update of configuration) requeues the object for reconcilation after the deletion timestamp is set. It's a little more messy than what I have above, but it should be possible to do using `Requeue` and `RequeueAfter` similarly.

Ultimately it will be good to add an integration test here to verify the behavior. Ideally the test follows this process:

1. marks an object for deletion several seconds into the future
2. verify that the controller logs include the `is set for deletion in the future` line indicating that it's seen the deletion and will wait
3. make an update to the `UDPRoute` that changes the configuration while it's marked for deletion
4. verify that the update succeeds and the dataplane gets the new configuration despite the fact that it's eventually going to get deleted (can be accomplished by checking logs, but also can be accomplished by [communicating directly with the dataplane](https://github.com/Kong/blixt/tree/main/internal/dataplane/client))
5. verify that the controller logs the `removing from dataplane` line indicating final removal
6. for extra thorough-ness use the [gRPC client](https://github.com/Kong/blixt/tree/main/internal/dataplane/client) to communicate with the dataplane and verify the full removal of the configuration data

Let me know if you have any questions, or how we can continue to help!

> **NOTE**: Don't forget that we'll need to add this behavior to both `UDPRoute` _and_ `TCPRoute` (feel free to generalize the functionality if you feel so inclined).

PR #86, opened 2023-05-02
shaneutt/dataplane tcp support

shaneutt:
- feat: add port tracking to BLIXT_CONNTRACK
- feat: track TCP connections
- feat: add TCP egress bpf program for dataplane
- chore: update ports for TCPRoute samples
- tests: add integration tests for TCPRoute

### DIFF
--- a/dataplane/ebpf/src/egress/tcp.rs
+++ b/dataplane/ebpf/src/egress/tcp.rs
@@ -1,0 +1,65 @@
+use core::mem;
+
+use aya_bpf::{
+    bindings::{TC_ACT_OK, TC_ACT_PIPE},
+    helpers::bpf_csum_diff,
+    programs::TcContext,
+};
+use aya_log_ebpf::info;
+
+use crate::{
+    bindings::{iphdr, tcphdr},
+    utils::{csum_fold_helper, ptr_at, ETH_HDR_LEN, IP_HDR_LEN},
+    BLIXT_CONNTRACK,
+};
+
+pub fn handle_tcp_egress(ctx: TcContext) -> Result<i32, i64> {
+    // gather the TCP header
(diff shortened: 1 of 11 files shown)


dataplane/ebpf/src/egress/tcp.rs
```diff
@@ -0,0 +1,65 @@
...
+        bpf_csum_diff(
+            mem::MaybeUninit::zeroed().assume_init(),
+            0,
+            ip_hdr as *mut u32,
+            mem::size_of::<iphdr>() as u32,
+            0,
+        )
+    } as u64;
+    unsafe { (*ip_hdr).check = csum_fold_helper(full_cksum) };
+    unsafe { (*tcp_hdr).check = 0 };
```
astoycos, line 60: Still not sure how this is working..... 

UDP  specifies 

`UDP checksum computation is optional for IPv4. If a checksum is not used it should be set to the value zero.`

But I didn't think TCP did

astoycos: Maybe because it's set to zero the kernel actually recalculates it...

astoycos (approved):
LGTM 

Thanks for this @shaneutt! 

I think the only thing I would add here is let's make an issue to also investigate the `tcp_hdr.check = 0` question to determine exactly how that's working.

dataplane/ebpf/src/egress/tcp.rs
shaneutt: Yeah there's some cleanup that needs to be done here, the scope of that cleanup is covered in https://github.com/Kong/blixt/issues/25 which should get everything consistent and checksums set up properly going forward. Given that you've also approved this PR, I'll take it that we're OK to continue that follow-up as adequate but do let me know if otherwise! :vulcan_salute:

PR #93, opened 2023-07-19
Add support for multiple gRPC client management

### LINKED ISSUE #46: Multiple grpc client endpoint management
## Problem Statement

For #41 we added a GRPC client in the controlplane which could connect to and configure the dataplane, but it requires to you create and manage a separate client for every dataplane endpoint (e.g. `Node`) that comes up. The purpose of this task is to create a higher level client that's capable of managing multiple dataplane connections and concurrently shipping new configuration to them all.

## Prerequisites

- [x] #41 

## Acceptance Criteria

- [ ] a single client object can be created which manages multiple dataplane endpoints (when present)
- [ ] the multi-endpoint client is capable of automatically adding new connections when new `Nodes` are added, and old `Nodes` are removed

levikobi:
Fixes #46.

### DIFF
--- a/controllers/dataplane_controller.go
+++ b/controllers/dataplane_controller.go
@@ -85,6 +85,11 @@ func (r *DataplaneReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&appsv1.DaemonSet{},
 			builder.WithPredicates(predicate.NewPredicateFuncs(r.daemonsetHasMatchingAnnotations)),
 		).
+		WithEventFilter(predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				return true
+			},
+		}).
 		Complete(r)
 }
 
--- a/controllers/dataplane_controller.go
+++ b/controllers/dataplane_controller.go
@@ -45,7 +45,7 @@ func NewDataplaneReconciler(client client.Client, schema *runtime.Scheme, manage
 		Client:                client,
 		Scheme:                schema,
 		BackendsClientManager: manager,
-		updates:               make(chan event.GenericEvent, 1000),
+		updates:               make(chan event.GenericEvent, 1),
 	}
 }
 
@@ -85,6 +85,7 @@ func (r *DataplaneReconciler) SetupWithManager(mgr ctrl.Manager) error {
 func (r *DataplaneReconciler) daemonsetHasMatchingAnnotations(obj client.Object) bool {
 	daemonset, ok := obj.(*appsv1.DaemonSet)
 	if !ok {
+		log.Error("received unexpected type in daemonset watch predicates: %T", obj)
 		return false
 	}
 
--- a/controllers/dataplane_controller.go
+++ b/controllers/dataplane_controller.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -33,9 +34,9 @@ import (
 // DataplaneReconciler reconciles the dataplane pods.
 type DataplaneReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	scheme *runtime.Scheme
 
-	BackendsClientManager *dataplane.BackendsClientManager
+	backendsClientManager *dataplane.BackendsClientManager
 
 	updates chan event.GenericEvent
 }
@@ -43,8 +44,8 @@ type DataplaneReconciler struct {
 func NewDataplaneReconciler(client client.Client, schema *runtime.Scheme, manager *dataplane.BackendsClientManager) *DataplaneReconciler {
 	return &DataplaneReconciler{
 		Client:                client,
-		Scheme:                schema,
-		BackendsClientManager: manager,
+		scheme:                schema,
+		backendsClientManager: manager,
 		updates:               make(chan event.GenericEvent, 1),
 	}
 }
@@ -57,6 +58,11 @@ var (
 // SetupWithManager loads the controller into the provided controller manager.
 func (r *DataplaneReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
+	// In order to allow our reconciler to quickly look up Pods by their owner, we’ll
+	// need an index. We declare an index key that we can later use with the client
+	// as a pseudo-field name, and then describe how to extract the indexed value from
+	// the Pod object. The indexer will automatically take care of namespaces for us,
+	// so we just have to extract the owner name if the Pod has a DaemonSet owner.
 	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &corev1.Pod{}, podOwnerKey, func(rawObj client.Object) []string {
 		// grab the pod object, extract the owner...
 		pod := rawObj.(*corev1.Pod)
@@ -83,9 +89,11 @@ func (r *DataplaneReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func (r *DataplaneReconciler) daemonsetHasMatchingAnnotations(obj client.Object) bool {
+	log := log.FromContext(context.Background())
+
 	daemonset, ok := obj.(*appsv1.DaemonSet)
 	if !ok {
-		log.Error("received unexpected type in daemonset watch predicates: %T", obj)
+		log.Error(fmt.Errorf("received unexpected type in daemonset watch predicates: %T", obj), "THIS SHOULD NEVER HAPPEN!")
 		return false
 	}
 
@@ -135,11 +143,15 @@ func (r *DataplaneReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	logger.Info("DataplaneReconciler", "reconcile status", "setting updated backends client list", "num ready pods", len(readyPodByNN))
-	updated, err := r.BackendsClientManager.SetClientsList(ctx, readyPodByNN)
+	updated, err := r.backendsClientManager.SetClientsList(ctx, readyPodByNN)
 	if updated {
 		logger.Info("DataplaneReconciler", "reconcile status", "backends client list updated, sending generic event")
-		r.updates <- event.GenericEvent{Object: ds}
-		logger.Info("DataplaneReconciler", "reconcile status", "generic event sent")
+		select {
+		case r.updates <- event.GenericEvent{Object: ds}:
+			logger.Info("DataplaneReconciler", "reconcile status", "generic event sent")
+		default:
+			logger.Info("DataplaneReconciler", "reconcile status", "generic event skipped - channel is full")
+		}
 	}
 	if err != nil {
 		logger.Error(err, "DataplaneReconciler", "reconcile status", "partial failure for backends client list update")
--- a/controllers/dataplane_controller.go
+++ b/controllers/dataplane_controller.go
@@ -1,0 +1,154 @@
+package controllers
+
+import (
+	"context"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	dataplane "github.com/kong/blixt/internal/dataplane/client"
+	"github.com/kong/blixt/pkg/vars"
+)
+
+//+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways/finalizers,verbs=update
+
+//+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=core,resources=services/status,verbs=get
+
+//+kubebuilder:rbac:groups=core,resources=endpoints,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=core,resources=endpoints/status,verbs=get
+
+// DataplaneReconciler reconciles the dataplane pods.
+type DataplaneReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+
+	BackendsClientManager *dataplane.BackendsClientManager
+
+	updates chan event.GenericEvent
+}
+
+func NewDataplaneReconciler(client client.Client, schema *runtime.Scheme, manager *dataplane.BackendsClientManager) *DataplaneReconciler {
+	return &DataplaneReconciler{
+		Client:                client,
+		Scheme:                schema,
+		BackendsClientManager: manager,
+		updates:               make(chan event.GenericEvent, 1000),
+	}
+}
+
+var (
+	podOwnerKey = ".metadata.controller"
+	apiGVStr    = appsv1.SchemeGroupVersion.String()
+)
+
+// SetupWithManager loads the controller into the provided controller manager.
+func (r *DataplaneReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &corev1.Pod{}, podOwnerKey, func(rawObj client.Object) []string {
+		// grab the pod object, extract the owner...
+		pod := rawObj.(*corev1.Pod)
+		owner := metav1.GetControllerOf(pod)
+		if owner == nil {
+			return nil
+		}
+		// ...make sure it's a DaemonSet...
+		if owner.APIVersion != apiGVStr || owner.Kind != "DaemonSet" {
+			return nil
+		}
+
+		// ...and if so, return it
+		return []string{owner.Name}
+	}); err != nil {
+		return err
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&appsv1.DaemonSet{},
+			builder.WithPredicates(predicate.NewPredicateFuncs(r.daemonsetHasMatchingAnnotations)),
+		).
+		Complete(r)
+}
+
+func (r *DataplaneReconciler) daemonsetHasMatchingAnnotations(obj client.Object) bool {
+	daemonset, ok := obj.(*appsv1.DaemonSet)
+	if !ok {
+		return false
+	}
+
+	// determine if this is a blixt daemonset
+	matchLabels := daemonset.Spec.Selector.MatchLabels
+	app, ok := matchLabels["app"]
+	if !ok || app != vars.DefaultDataPlaneAppLabel {
+		return false
+	}
+
+	// verify that it's the dataplane daemonset
+	component, ok := matchLabels["component"]
+	if !ok || component != vars.DefaultDataPlaneComponentLabel {
+		return false
+	}
+
+	return true
+}
+
+// Reconcile provisions (and de-provisions) resources relevant to this controller.
+func (r *DataplaneReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	ds := new(appsv1.DaemonSet)
+	if err := r.Client.Get(ctx, req.NamespacedName, ds); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("DataplaneReconciler", "reconcile status", "object enqueued no longer exists, skipping")
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	var childPods corev1.PodList
+	if err := r.List(ctx, &childPods, client.InNamespace(req.Namespace), client.MatchingFields{podOwnerKey: req.Name}); err != nil {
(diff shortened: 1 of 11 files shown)


shaneutt:
Awesome! Thanks for putting in the effort on this one, let's run CI and see how we do.

Also please do check out https://github.com/Kong/blixt/discussions/42#discussioncomment-6490256, we should be really close to completing our donation to the CNCF and we're trying to document CNCF CLA completion for all contributors as a requirement.

shaneutt (commented):
Thanks for jumping in to do this one @levikobi :partying_face: 

One thing I noticed while reviewing is that we've added a bit of machinery of our own to reconcile things, e.t.c the `observer` interface and the `SetupReconcilation` method as opposed to leaning into the `controller-runtime` machinery more as we had been doing previously. Could you tell me more about the decisions made here, if there were difficulties with trying to rely more on the traditional reconciliation approach?

levikobi:
> Thanks for jumping in to do this one @levikobi 🥳
> 
> One thing I noticed while reviewing is that we've added a bit of machinery of our own to reconcile things, e.t.c the `observer` interface and the `SetupReconcilation` method as opposed to leaning into the `controller-runtime` machinery more as we had been doing previously. Could you tell me more about the decisions made here, if there were difficulties with trying to rely more on the traditional reconciliation approach?

Thank you for letting me get involved @shaneutt.

Regarding your question, this is my first interaction with the controller-runtime SDK, so I tried doing my best while also keeping in mind that this change might not stay for too long.

I do understand from your comment though that I could have made a better usage out of the controller runtime.😅

If you can give me some tips about improving things, I will happily apply them. Anyway, I will try to go over the documentation again and see what I've missed.

Thanks :)

shaneutt:
> > Thanks for jumping in to do this one @levikobi partying_face
> > One thing I noticed while reviewing is that we've added a bit of machinery of our own to reconcile things, e.t.c the `observer` interface and the `SetupReconcilation` method as opposed to leaning into the `controller-runtime` machinery more as we had been doing previously. Could you tell me more about the decisions made here, if there were difficulties with trying to rely more on the traditional reconciliation approach?
> 
> Thank you for letting me get involved @shaneutt.
> 
> Regarding your question, this is my first interaction with the controller-runtime SDK, so I tried doing my best while also keeping in mind that this change might not stay for too long.
> 
> I do understand from your comment though that I could have made a better usage out of the controller runtime.sweat_smile
> 
> If you can give me some tips about improving things, I will happily apply them. Anyway, I will try to go over the documentation again and see what I've missed.
> 
> Thanks :)

We really appreciate you jumping in here!

We use [operator-sdk](https://sdk.operatorframework.io/build/) (which uses [kubebuilder](https://book.kubebuilder.io/) under the hood, which then uses [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime) under the hood) and you can have that generate controllers for you that will provide a standard `Reconcile()` callback and load it into your controller manager.

I wouldn't say we're strictly opposed to the solution you've put forth here, but I do think it changes how our controllers work in a way that handles more of the reconciliation machinery more manually than I would have expected. If we continued using our previous pattern I think we could just add one new controller that watches the `DaemonSet` for the containers to change and update the dataplane client with any new changes. Overall:

1. make the dataplane client support multiple backends
2. provide threadsafe update functionality for the dataplane client
3. add a new controller which responds changes in `Pods` owned by the `Daemonset` and updates the dataplane client accordingly
4. ensure updates to the dataplane client's backends result in triggering a new configuration push

I think ideally we would try to do all this within the confines of the higher level mechanisms (reconciler callback types, and watches), though I admit number 4 is the thing that makes this hard to do. I think I would have to experiment a bit with the various watch sources provided by controller-runtime: using a channel source seems like it could be one reasonably simple way to do it, all dataplane client updates resulting in a channel message that triggers re-reconcilation of `TCPRoutes`/`UDPRoutes`, though that has some efficiency issues and goes a bit against the grain of what the channel source was meant for.

How about this: do some thinking and looking around let me know if anything stands out to you, and I'll do the same plus also do another round of review here and see if I can come up with anything. Ultimately I don't want you to feel like "you stepped in it" with this one like you're getting pulled into more than you bargained for, so just know that I'm prioritizing your time as a contributor as well as working through some maintenance concerns.

levikobi:
Thanks for the honest feedback @shaneutt! I understand.
I will finish my prior commitments at Gateway-API and ingress2gateway before the release date and come back to this one afterward to make things right.

main.go
```diff
@@ -75,6 +78,21 @@ func main() {
 		os.Exit(1)
 	}
 
+	clientsManager, err := client.NewBackendsClientManager(cfg)
+	if err != nil {
+		setupLog.Error(err, "unable to create backends client manager")
```
shaneutt, line 83:
```suggestion
		setupLog.Error(err, "unable to create backends client manager")
		os.Exit(1)
```

main.go
```diff
@@ -115,8 +137,60 @@ func main() {
...
 	setupLog.Info("starting manager")
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+	if err := mgr.Start(ctx); err != nil {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
 }
+
+// Tee consumes the received channel and mirrors the messages into 2 new channels.
+func Tee[T any](ctx context.Context, in <-chan T) (_, _ <-chan T) {
```
shaneutt, line 147: This fanout helper doesn't need to be public:

```suggestion
func tee[T any](ctx context.Context, in <-chan T) (_, _ <-chan T) {
```

main.go
```diff
@@ -115,8 +137,60 @@ func main() {
...
+			case out2 <- val:
+				select {
+				case <-ctx.Done():
+				case out1 <- val:
+				}
+			}
+		}
+	}()
+	return out1, out2
+}
```
shaneutt, line 196: Question: was this fanout function something you put together just for this PR, or was it used from elsewhere? If it was taken from elsewhere are there some unit tests that could come along for the ride?

shaneutt (commented):
Did a first pass of review, mostly I think we're looking pretty solid. I haven't dug deep into the route controllers yet. Most of my comments so far are minor, I do think the channel approach should be a workable solution thank you for putting the time and energy into this we really appreciate it! I'll try to get to my next pass soon.

main.go
levikobi: Agree. I think it can't be used anyways since we'll end up with a dependency cycle

controllers/dataplane_controller.go
```diff
@@ -0,0 +1,154 @@
...
+
+//+kubebuilder:rbac:groups=core,resources=endpoints,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=core,resources=endpoints/status,verbs=get
+
+// DataplaneReconciler reconciles the dataplane pods.
+type DataplaneReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+
+	BackendsClientManager *dataplane.BackendsClientManager
```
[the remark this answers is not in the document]
levikobi: Since I ended up adding a constructor for this struct you're right, it doesn't need to be public. I'll change it.

controllers/dataplane_controller.go
```diff
@@ -0,0 +1,154 @@
...
+		// ...make sure it's a DaemonSet...
+		if owner.APIVersion != apiGVStr || owner.Kind != "DaemonSet" {
+			return nil
+		}
+
+		// ...and if so, return it
+		return []string{owner.Name}
+	}); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
levikobi: You're right, I probably wouldn't get it either without reading the docs.
Took the description from Kubebuilder's book

main.go
levikobi: This was taken from a library I've written, containing common concurrency patterns for channels (heavily inspired by Katherine Cox-Buday's book - Concurrency in Go).

I don't have unit tests for it, but I've been using it for a while so I felt comfortable adding it..
Let me know if unit tests are important for you here, I'll figure something out

controllers/dataplane_controller.go
```diff
@@ -0,0 +1,154 @@
...
+
+	updates chan event.GenericEvent
+}
+
+func NewDataplaneReconciler(client client.Client, schema *runtime.Scheme, manager *dataplane.BackendsClientManager) *DataplaneReconciler {
+	return &DataplaneReconciler{
+		Client:                client,
+		Scheme:                schema,
+		BackendsClientManager: manager,
+		updates:               make(chan event.GenericEvent, 1000),
```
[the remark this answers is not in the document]
shaneutt: Yeah I think this should work :thinking: 

At this point in the code if we are guaranteed to have the backends updated in the client, then it seems to follow that we should be able to "collapse" stacked up events by simply dropping extras, since all messages mean the same thing and trigger the same update. Let's give that a try :+1:

controllers/dataplane_controller.go
```diff
@@ -0,0 +1,154 @@
...
+				readyPodByNN[key] = pod
+			}
+		}
+	}
+
+	logger.Info("DataplaneReconciler", "reconcile status", "setting updated backends client list", "num ready pods", len(readyPodByNN))
+	updated, err := r.BackendsClientManager.SetClientsList(ctx, readyPodByNN)
+	if updated {
+		logger.Info("DataplaneReconciler", "reconcile status", "backends client list updated, sending generic event")
+		r.updates <- event.GenericEvent{Object: ds}
```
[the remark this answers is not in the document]
shaneutt: OR we can try "collapsing" additional messages as you suggested above.

main.go
shaneutt: Ah yeah, the book with the shrew on it right? I have that one.

I did some experimenting locally and I do think we could probably approach this a few different ways, but in retrospect given the grand scheme of things we probably don't need to focus on this too much for this PR, it will be very easy to iterate on this as needed in follow-up PRs so we can keep this PR more focused on the task at hand.

controllers/dataplane_controller.go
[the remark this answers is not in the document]
levikobi: Awesome!

main.go
levikobi: > Ah yeah, the book with the shrew on it right? I have that one.

Yes exactly! It's a nice one

> I did some experimenting locally and I do think we could probably approach this a few different ways, but in retrospect given the grand scheme of things we probably don't need to focus on this too much for this PR

Totally agree, it was just the fastest and simplest solution I could come up with (given that I had this functionality from before lol).

> It will be very easy to iterate on this as needed in follow-up PRs so we can keep this PR more focused on the task at hand.

👍

internal/dataplane/client/client.go
```diff
@@ -2,64 +2,212 @@ package client
...
+	return &BackendsClientManager{
+		log:       log.FromContext(context.Background()),
+		clientset: clientset,
+		mu:        sync.RWMutex{},
+		clients:   map[types.NamespacedName]clientInfo{},
+	}, nil
+}
+
+func (c *BackendsClientManager) SetClientsList(ctx context.Context, readyPods map[types.NamespacedName]corev1.Pod) (bool, error) {
+	// TODO: close and connect to the different clients concurrently.
```
shaneutt, line 55: We need to resolve this TODO in one of two ways:

1. implement the solution now, as a part of this PR
2. if we can live without the solution temporarily, create a follow-up issue that explains what is needed, and why, and what the downsides are of not having it, and link it here in the comment so that we don't lose track of this need.

shaneutt (commented):
I think this is close to being ready, just had a comment about a TODO we should resolve or follow-up on. Also @astoycos should take a look too since this is a fairly large change and two heads are better than one.

At a higher level though, it seems we're missing an integration/e2e style test that validates these changes. One way we might do this is make the default `kind` cluster that's created for the tests include 3 nodes instead of 1, but by default cordon off 2/3 of them.

Our testing framework provides a `WithConfig()` option for the `kind` cluster builder:

https://github.com/Kong/kubernetes-testing-framework/blob/main/pkg/clusters/types/kind/builder.go#L54

I think it might be reasonable for us to provide a 3-node configuration to that in an e2e test that then cordons off the nodes, then removes the cordon to instigate the `Daemonset` deleting, and then creating the pods, and then we could verify the configuration (for `TCPRoute`/`UDPRoute`) gets applied to the `Pods` as they come up, _and_ we could test the traffic and make sure things are working as expected?

Let me know your thoughts on this @levikobi, and if that sounds like a lot or if it sounds pretty straightforward? Curious if you have any other thoughts on testing as well. :thinking:

shaneutt:
There are also some lint issues that need to be addressed, according to CI:

```console
  Error: : # github.com/kong/blixt
  Error: ./main.go:95:54: undefined: Tee (typecheck)
```

controllers/dataplane_controller.go
```diff
@@ -0,0 +1,167 @@
...
+
+		// ...and if so, return it
+		return []string{owner.Name}
+	}); err != nil {
+		return err
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&appsv1.DaemonSet{},
+			builder.WithPredicates(predicate.NewPredicateFuncs(r.daemonsetHasMatchingAnnotations)),
```
astoycos, line 86: Can you possibly filter this even more? i.e Only care about events where the DS's status changes since it's where we get the ready pods from anyways? 

```
status:
  currentNumberScheduled: 1
  desiredNumberScheduled: 1
  numberAvailable: 1
  numberMisscheduled: 0
  numberReady: 1
  observedGeneration: 1
  updatedNumberScheduled: 1
```

main.go
```diff
@@ -115,8 +138,60 @@ func main() {
...
 	setupLog.Info("starting manager")
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+	if err := mgr.Start(ctx); err != nil {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
 }
+
+// Tee consumes the received channel and mirrors the messages into 2 new channels.
+func tee[T any](ctx context.Context, in <-chan T) (_, _ <-chan T) {
```
astoycos, line 148: Whiy does this function need to be generic? The type we're sending over the update channels is always `event.GenericEvent` right?

main.go
```diff
@@ -115,8 +138,60 @@ func main() {
...
+	}
+
+	go func() {
+		defer close(out1)
+		defer close(out2)
+
+		for val := range OrDone(ctx, in) {
+			select {
+			case <-ctx.Done():
+
```
astoycos, line 181: Are we supposed to return here?

controllers/udproute_controller.go
```diff
@@ -35,7 +36,9 @@ type UDPRouteReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
 
-	log logr.Logger
+	log                   logr.Logger
+	ReconcileRequestChan  <-chan event.GenericEvent
```
astoycos, line 40: NIT can you rename this to like "ClientReconcileRequestChan" Or somthing? I just want it to be well known that this infra IS NOT provided by operator-sdk :)

controllers/tcproute_controller.go
```diff
@@ -35,7 +36,9 @@ type TCPRouteReconciler struct {
 	client.Client
```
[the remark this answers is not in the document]
astoycos: Can we make sure that all controller specific RBAC tags are aligned to each dedicated controller file and then any RBAC tags that are used in all of the controllers maybe goes in the dataplane. 

Sorry I know it wasn't really a problem you introduced :/

controllers/tcproute_controller.go
```diff
@@ -208,19 +204,12 @@ func (r *TCPRouteReconciler) ensureTCPRouteDeletedInDataPlane(ctx context.Contex
...
-
 	// delete the target from the dataplane
-	confirmation, err := dataplaneClient.Delete(context.Background(), targets.Vip)
-	if err != nil {
+	if _, err = r.BackendsClientManager.Delete(ctx, targets.Vip); err != nil {
 		return err
 	}
 
-	r.log.Info(fmt.Sprintf("successful data-plane DELETE, confirmation: %s", confirmation.String()))
+	r.log.Info("successful data-plane DELETE")
```
astoycos, line 212: nit any reason not to print the confirmation still here?

astoycos (commented):
Awesome work @levikobi, thanks so much for tackling this :) 

I generally have a bunch of nits and a few questions otherwise I think this is a pretty cool design 👍.  Lets just make sure we try and keep as clear of a separation as possible between operator-sdk enabled funcations/vars/uses and the custom channel logic you've added here with comments and smart variable naming. I just push this to not confuse folks in the future.

Additionally we are adding a little complexity so I'd like to make sure we have at least some unit/integration testing added to make sure all of the channel functionality is working as expected. Feel free to make an issue for this and tackle it in a follow up or ask someone else to :)

main.go
levikobi: It doesn't have to be generic, I just thought it would be nicer this way.
As you said, we're only using it for `event.GenericEvent` so if you want I can switch it up to the concrete type, please let me know what you prefer

main.go
levikobi: Yes. We will return because the channel returned from `OrDone` is going to be closed, hence the loop at 178 will end.
I probably should add a return statement at line 181 for clarity though

controllers/udproute_controller.go
levikobi: Sure, done

controllers/tcproute_controller.go
levikobi: I just think it will be more informative to print the information about each pod whenever each individual request ends, instead of waiting for them all and then aggregating everything together.

Let me know if you prefer it the other way around though

controllers/dataplane_controller.go
levikobi: I'm still relatively new to the tooling, did you mean something like that?
```go
	return ctrl.NewControllerManagedBy(mgr).
		For(&appsv1.DaemonSet{},
			builder.WithPredicates(predicate.NewPredicateFuncs(r.daemonsetHasMatchingAnnotations)),
		).
		WithEventFilter(predicate.Funcs{
			UpdateFunc: func(e event.UpdateEvent) bool {
				return true
			},
		}).
		Complete(r)
```

levikobi:
> So keeping in mind that this functionality is not meant to exist long term, I think this is close to being ready and we can move forward with this. I added a comment about a TODO we should resolve or follow-up on. Also @astoycos should take a look too since this is a fairly large change (and there's multiple ways we could approach this) and two heads are better than one.
At a higher level though, it seems we're missing an integration/e2e style test that validates these changes. One way we might do this is make the default kind cluster that's created for the tests include 3 nodes instead of 1, but by default cordon off 2/3 of them.
Our testing framework provides a WithConfig() option for the kind cluster builder:
https://github.com/Kong/kubernetes-testing-framework/blob/main/pkg/clusters/types/kind/builder.go#L54
I think it might be reasonable for us to provide a 3-node configuration to that in an e2e test that then cordons off the nodes, then removes the cordon to instigate the Daemonset deleting, and then creating the pods, and then we could verify the configuration (for TCPRoute/UDPRoute) gets applied to the Pods as they come up, and we could test the traffic and make sure things are working as expected?
Let me know your thoughts on this @levikobi, and if that sounds like a lot or if it sounds pretty straightforward? Curious if you have any other thoughts on testing as well. 🤔

Sound like a good idea, testing is definitely a must for this change.
I'll try to get it done and push it probably during next week

levikobi:
> Awesome work @levikobi, thanks so much for tackling this :)
> 
> I generally have a bunch of nits and a few questions otherwise I think this is a pretty cool design 👍. Lets just make sure we try and keep as clear of a separation as possible between operator-sdk enabled funcations/vars/uses and the custom channel logic you've added here with comments and smart variable naming. I just push this to not confuse folks in the future.
> 
> Additionally we are adding a little complexity so I'd like to make sure we have at least some unit/integration testing added to make sure all of the channel functionality is working as expected. Feel free to make an issue for this and tackle it in a follow up or ask someone else to :)

Thanks for the review @astoycos!

controllers/dataplane_controller.go
astoycos: Yeah close to that see https://sdk.operatorframework.io/docs/building-operators/golang/references/event-filtering/

In bpfd we do something like 

```
// Only reconcile if a bpfprogram has been created for the controller's program type.
func BpfProgramTypePredicate(kind string) predicate.Funcs {
	return predicate.Funcs{
		GenericFunc: func(e event.GenericEvent) bool {
			return e.Object.(*bpfdiov1alpha1.BpfProgram).Spec.Type == kind
		},
		CreateFunc: func(e event.CreateEvent) bool {
			return e.Object.(*bpfdiov1alpha1.BpfProgram).Spec.Type == kind
		},
		UpdateFunc: func(e event.UpdateEvent) bool {
			return e.ObjectNew.(*bpfdiov1alpha1.BpfProgram).Spec.Type == kind
		},
		DeleteFunc: func(e event.DeleteEvent) bool {
			return e.Object.(*bpfdiov1alpha1.BpfProgram).Spec.Type == kind
		},
	}
}
```

astoycos: (which is a bit verbose but has worked for us in the past)

main.go
astoycos: I would prefer the concrete type unless @shaneutt thinks otherwise :) 

Generics are super super cool but I'd expect we'll end up with alot of Go newbies on this project

controllers/tcproute_controller.go
astoycos: ++ agreed

shaneutt:
> Sound like a good idea, testing is definitely a must for this change. I'll try to get it done and push it probably during next week

Ok, reach out to me if you need any help with this there's some room for interpretation here, it's not terribly straightforward.

main.go
shaneutt: I don't feel strongly about this, @astoycos seems to feel a little bit more strongly about it than I do, so unless you're feeling very strongly it should remain generic I would say let's just make it explicit like Andrew suggests since there's really no downside of that.

levikobi: Sure, no problem

shaneutt:
Hey! :wave: 

Just checking in. Need any help to move this one forward?

internal/dataplane/client/client.go
levikobi: I've opened https://github.com/kubernetes-sigs/blixt/issues/127 to address the issue

shaneutt (approved):
Talked with @astoycos and rather than hold this PR up and making it bigger (since it's already pretty good size) we're aligned on taking what we have here, and following up (thanks for creating follow-ups like #127).

So we'll follow up with some tweaks and tests, but so far integration is passing, thank you for all your effort on this @levikobi! :vulcan_salute: 

/approve
/lgtm

Issue #95, opened 2023-08-23
Gateway API - initial conformance with GatewayClass and Gateway resources


shaneutt:
As it relates to #81, we want to plug in the conformance test suite from [Gateway API](https://github.com/kubernetes-sigs/gateway-api) into Blixt as the first step before #81 can be resolved.

## Acceptance Criteria

- [x] a `test/conformance` test suite is added which runs the Gateway API conformance test suite
- [x] at least some, if not all of the `GatewayClass` and `Gateway` tests are run
- [x] CI is configured to run the conformance test suite for every PR

shaneutt:
Resolved by #103

PR #103, opened 2023-09-15
Initial GatewayClass/Gateway Conformance Setup

shaneutt:
This patch adds the initial scaffolding to start running the Gateway API conformance test suite on every PR.

This includes _some_ Gateway support related testing, but some tests are disabled for the moment as some are broken in upstream, and some specification simply isn't implemented yet here in Blixt.

This doesn't resolve, but it does support #95.


astoycos (approved):
LGTM

thanks @shaneutt

PR #117, opened 2023-10-15
udproute: add support for round-robin load balancing

### LINKED ISSUE #10: UDP round-robin load-balancing
## Problem Statement

Previously we've added the ability to redirect UDP traffic to a `Backend` using eBPF, but only to a single `Backend`. The purpose of this task is to add support for multiple `Backends` in the `Routing` map, and then support to balance traffic between those `Backends` using the "round-robin" strategy in our eBPF data-plane code.

## Prerequisites

- [x] https://github.com/Kong/blixt/issues/8
- [x] https://github.com/Kong/blixt/issues/9

## Proposed Solution

- [ ] the `BACKENDS` map can now include multiple `BACKEND` resources
- [ ] the eBPF code is updated to handle load-balancing between `BACKEND` resources according to a round-robin strategy
- [ ] integration tests are added that provide a `UDPRoute` pointing to a `Service` results in the correct dispersion of requests across multiple `BACKEND` resources
- [ ] each `UDPRoute` should be able to support multiple `BackendRefs`
- [ ] each `BackendRef` for a `UDPRoute` should be able to support multiple `Endpoints`

## Acceptance Criteria

- [ ] when a `Service` is used as a `backendRef` for a `UDPRoute` and it includes multiple endpoints, traffic is load-balanced between those endpoints with a basic round-robin strategy (future iterations may add other strategies, but not required to resolve this).

aryan9600:
Add support for distributing traffic amongst several backends for a UDPRoute in a round-robin fashion. High level changes:
* The `target` field in the `Targets` protobuf message has been updated to be a list named `targets`.
* The bpf map `BACKENDS` has been updated to be of type `<BackendKey, BackendList>`, where `BackendList` contains a list of `Backends`.
* A new bpf map `GATEWAY_INDEXES` of type `<BackendKey, u16>` has been introduced to keep track of the last backend that received a packet.

Deletion handling for TCPRoute and UDPRoute has also been improved by avoiding fetching related `Endpoints`, which could lead to objects being stuck with finalizers.

Fixes #10


dataplane/ebpf/src/main.rs
```diff
@@ -36,8 +36,12 @@ use utils::{ETH_HDR_LEN, ETH_P_IP, IPPROTO_ICMP, IPPROTO_TCP, IPPROTO_UDP};
 // -----------------------------------------------------------------------------
 
 #[map(name = "BACKENDS")]
-static mut BACKENDS: HashMap<BackendKey, Backend> =
-    HashMap::<BackendKey, Backend>::with_max_entries(128, 0);
+static mut BACKENDS: HashMap<BackendKey, BackendList> =
+    HashMap::<BackendKey, BackendList>::with_max_entries(128, 0);
```
shaneutt, line 40: Let's define another constant for this, that way we can just change it once if we ever need to change it, e.t.c.

dataplane/xtask/src/grpc.rs
```diff
@@ -49,11 +49,11 @@ pub async fn update(opts: Options) -> Result<(), Error> {
 
     let req = Request::new(Targets {
         vip: Some(vip.clone()),
-        target: Some(Target {
+        targets: vec![Target {
             daddr: daddr.into(),
             dport: opts.dport,
             ifindex: Some(opts.ifindex),
-        }),
+        }],
```
shaneutt, line 56: We talked about this one in Zoom, in this xtask we're populating some "test data" to play around with, it could be helpful to add multiple targets here for testing the new load-balancing behavior. Doesn't necessarily have to happen as a part of this PR though if you're strapped for time, just add a comment/followup.

internal/dataplane/client/utils.go
```diff
@@ -31,60 +31,57 @@ import (
...
+				}
+
+				for _, addr := range subset.Addresses {
+					if addr.IP == "" {
+						return nil, fmt.Errorf("empty IP for endpoint subset")
+					}
+
+					ip := net.ParseIP(addr.IP)
+
+					podip := binary.BigEndian.Uint32(ip.To4())
```
shaneutt, line 67: There's a couple places where we implicitly assume V4, this is totally fine for now, but if we don't have an issue tracking adding V6 support we should create that.

internal/dataplane/client/utils.go
```diff
@@ -166,7 +163,7 @@ func CompileTCPRouteToDataPlaneBackend(ctx context.Context, c client.Client, tcp
 			Ip:   ipint,
 			Port: gatewayPort,
 		},
-		Target: target,
+		Targets: []*Target{target},
```
shaneutt, line 166: Similar to above where we said we would do TCPRoute later, maybe just link the relevant issue here in a TODO.

shaneutt (commented):
This is looking really good, thank you for taking the time to knock this one out @aryan9600!

This review was my first pass that you, @astoycos and myself jumped on a Zoom for, and most of these comments are related to follow ups e.t.c.

We'll want to make sure we get review from Andrew as well:

/cc @astoycos

internal/dataplane/client/utils.go
astoycos: https://github.com/kubernetes-sigs/blixt/issues/120

Issue #106, opened 2023-10-11
Docker build CI needs an update post kubernetes-sigs migration


shaneutt:
Now that we're migrated from https://github.com/kong/blixt to https://github.com/kubernetes-sigs/blixt, some updates need to be made to our container image builds so that we can build and push our images in CI.

ameukam:
GHCR is currently not advised for use since we don't have clarity how repositories exposure. SIG k8s Infra always recommend to use GCP GCR/AR repos for all the subprojects. We have 2 options: 

- setup a dedicated GCR repo
- use an existing one owned by SIG Network: https://github.com/kubernetes/k8s.io/blob/main/k8s.gcr.io/images/k8s-staging-networking/images.yaml

shaneutt:
Personally I think using an existing one for now is fine, especially if that's easier / more straightforward.

ameukam:
> Personally I think using an existing one for now is fine, especially if that's easier / more straightforward.

With k8s-infra hat on, the first option is preferable with isolation of resource at the project and minimize any security issue. I'll try to setup it if we go with this approach.

shaneutt:
I'm in favor. @aryan9600, @astoycos, @mlavacca ?

aryan9600:
sounds good to me!

PR #121, opened 2023-10-20
Bpfd integration

astoycos:
This sets up the integration with [bpfd](bpfd.dev)

Additionally it adds a deployment scenario to the integration tests so that we run with and without bpfd.

To test it out locally 
```
1. Build Cluster: make build.cluster
2. Install bpfd: kubectl apply -k ./config/bpfd-install
3. Install Blixt with bpfd programs: kubectl apply -k ./config/bpfd
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,6 +188,10 @@ ifndef ignore-not-found
   ignore-not-found = false
 endif
 
+.PHONY: install-bpfd
+install-bpfd: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
+	$(KUSTOMIZE) build config/bpfd | kubectl apply -f -
+
 .PHONY: install
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/crd | kubectl apply -f -
@@ -315,4 +319,11 @@ build.cluster: $(KTF) # builds a KIND cluster which can be used for testing and 
 .PHONY: load.image
 load.image: build.image
 	kind load docker-image $(BLIXT_CONTROLPLANE_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
+		kubectl -n blixt-system rollout restart deployment blixt-controlplane
+
+.PHONY: load.all.images
+load.all.images: build.all.images
+	kind load docker-image $(BLIXT_CONTROLPLANE_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
+	kind load docker-image $(BLIXT_DATAPLANE_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
+	kind load docker-image $(BLIXT_UDP_SERVER_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
 		kubectl -n blixt-system rollout restart deployment blixt-controlplane
--- a/config/bpf-bytecode/tc-egress.yaml
+++ b/config/bpf-bytecode/tc-egress.yaml
@@ -1,0 +1,21 @@
+apiVersion: bpfd.dev/v1alpha1
+kind: TcProgram
+metadata:
+  labels:
+    app.kubernetes.io/name: blixt-tc-egress
+  name: tc-egress
+spec:
+  bpffunctionname: tc_egress
+  # Select all nodes
+  nodeselector: {}
+  interfaceselector:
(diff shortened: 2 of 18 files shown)


astoycos:
TODO: I need follow up issues for

1. Building/pushing blixt bytecode images in CI
2. Pointing back to bpfd:main rather than bpfd:blixt for the bpfd install images
3. Adding documentation

Makefile
```diff
@@ -316,3 +320,10 @@ build.cluster: $(KTF) # builds a KIND cluster which can be used for testing and
 load.image: build.image
 	kind load docker-image $(BLIXT_CONTROLPLANE_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
 		kubectl -n blixt-system rollout restart deployment blixt-controlplane
+
+.PHONY: load.all.images
+load.all.images: build.all.images
+	kind load docker-image $(BLIXT_CONTROLPLANE_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
+	kind load docker-image $(BLIXT_DATAPLANE_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
+	kind load docker-image $(BLIXT_UDP_SERVER_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
+		kubectl -n blixt-system rollout restart deployment blixt-controlplane
```
shaneutt, line 329: :+1:

config/bpf-bytecode/tc-egress.yaml
```diff
@@ -0,0 +1,21 @@
...
+  bpffunctionname: tc_egress
+  # Select all nodes
+  nodeselector: {}
+  interfaceselector:
+    primarynodeinterface: true
+  priority: 0
+  direction: egress
+  bytecode:
+    image:
+      url: quay.io/bpfd-bytecode/blixt-tc-egress:latest
```
shaneutt, line 17: Should we put a TODO here to get this image pushed to the same place we push our other images?

(note that we have #106 to deal with too, I hadn't gotten around to it yet)

config/bpfd/bpfd.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: apps/v1
```
shaneutt, line 1: We should probably add some comments with an explanation of what's going on with this `DaemonSet`, right?

dataplane/.cargo/config.toml
```diff
@@ -1,2 +1,6 @@
 [alias]
-xtask = "run --package xtask --"
\ No newline at end of file
+xtask = "run --package xtask --"
+
+## Needed to build blixt's programs with BTF + custom bpf-linker build.
```
shaneutt, line 4: :+1:

dataplane/Makefile
```diff
@@ -26,3 +26,25 @@ build.image:
...
+      --build-arg BPF_FUNCTION_NAME=tc_egress \
+      --build-arg PROGRAM_TYPE=tc \
+      --build-arg BYTECODE_FILENAME=loader \
+      -f https://raw.githubusercontent.com/bpfd-dev/bpfd/main/packaging/container-deployment/Containerfile.bytecode \
+	  ./target/bpfel-unknown-none/debug -t quay.io/bpfd-bytecode/blixt-tc-egress:latest
+
+.PHONY: push.bytecode.images
+push.bytecode.images: build.bytecode.images
+	docker push quay.io/bpfd-bytecode/blixt-tc-egress:latest
+	docker push quay.io/bpfd-bytecode/blixt-tc-ingress:latest
```
shaneutt, line 50: Similar to above, I guess we just may want to consolidate our images under one room, or at the very least get this set up so that CI can actually push these because I assume right now it has to be done manually?

At minimum I think we should just spin off an issue and link it here pointing out that we need to do some maintenance regarding where images are pushed, could perhaps just add the context to #106

dataplane/ebpf/.cargo/config.toml
```diff
@@ -1,6 +1,7 @@
 [build]
 target-dir = "../target"
 target = "bpfel-unknown-none"
+rustflags = ["-C", "save-temps"]
```
shaneutt, line 4: I think it would be nice to add comments here about our flags for future readers

shaneutt (commented):
Mostly looking good, glad to see us adding bpfd!

Mostly small comments.

shaneutt (approved):
/approve
/lgtm

PR #148, opened 2023-12-01
tcproute: add support for round-robin load balancing

### LINKED ISSUE #119: Add support for load balancing TCP backends
## Problem Statement

At the moment, support for TCP backends is more int the form of a proxy than a  load balancer, i.e. it can handle only one TCP backend at a time. We should add support for load balancing traffic between multiple TCP backends in a round robin fashion. This will allow users to specify more than one backend reference in a `TCPRoute` object.

## Prerequisites

- [x] #10

aryan9600:
Add support for round-robin load balancing for `TCPRoute`. This enables mutliple backend references in a UDPRoute object, with traffic being distributed to each backend in a round-robin fashion.

A new BPF map `TCP_CONNECTIONS` was introduced to help keep track of active TCP connections, with its key storing the clien <<ip:port>> identifier and the value storing the backend's <<ip:port>> along with the state of the connection.

Fixes #119 

Follow up items:
* modify `TCP_CONNECTIONS` to be more generic; get rid of `BLIXT_CONNTRACK` and refactor udp ingress and icmp egress to use the new generic map

### DIFF
--- a/config/tests/tcproute-rr/patch.yaml
+++ b/config/tests/tcproute-rr/patch.yaml
@@ -1,0 +1,17 @@
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: TCPRoute
+metadata:
+  name: blixt-tcproute-sample
+spec:
+  parentRefs:
+  - name: blixt-tcproute-sample
+    port: 8080
+  rules:
+  - backendRefs:
+    - name: blixt-tcproute-sample
+      port: 8080
+  - backendRefs:
+    - name: tcproute-rr-v1
+      port: 8080
+    - name: tcproute-rr-v2
+      port: 8080
--- a/config/tests/tcproute-rr/server.yaml
+++ b/config/tests/tcproute-rr/server.yaml
@@ -1,0 +1,76 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: tcproute-rr-v1
+  labels:
+    app: tcproute-rr-v1
+spec:
+  selector:
+    matchLabels:
+      app: tcproute-rr-v1
+  template:
+    metadata:
+      labels:
+        app: tcproute-rr-v1
+    spec:
+      containers:
+      - name: tcp-echo
+        image: istio/tcp-echo-server:1.1
+        imagePullPolicy: IfNotPresent
+        args: [ "8080", "tcproute-rr-v1:" ]
+        ports:
+        - containerPort: 8080
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: tcproute-rr-v2
+  labels:
+    app: tcproute-rr-v2
+spec:
+  selector:
+    matchLabels:
+      app: tcproute-rr-v2
+  template:
+    metadata:
+      labels:
+        app: tcproute-rr-v2
+    spec:
+      containers:
+      - name: tcp-echo
+        image: istio/tcp-echo-server:1.1
+        imagePullPolicy: IfNotPresent
+        args: [ "8080", "tcproute-rr-v2:" ]
+        ports:
+        - containerPort: 8080
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: tcproute-rr-v1
+  name: tcproute-rr-v1
+spec:
+  ports:
+  - name: tcp
+    port: 8080
+    protocol: TCP
+  selector:
+    app: tcproute-rr-v1
+  type: ClusterIP
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: tcproute-rr-v2
+  name: tcproute-rr-v2
+spec:
+  ports:
+  - name: tcp
+    port: 8080
+    protocol: TCP
+  selector:
+    app: tcproute-rr-v2
+  type: ClusterIP
--- a/dataplane/api-server/src/server.rs
+++ b/dataplane/api-server/src/server.rs
@@ -8,28 +8,31 @@ use std::net::Ipv4Addr;
 use std::sync::Arc;
 
 use anyhow::Error;
-use aya::maps::{HashMap, MapData};
+use aya::maps::{HashMap, MapData, MapError};
 use tokio::sync::Mutex;
 use tonic::{Request, Response, Status};
 
 use crate::backends::backends_server::Backends;
 use crate::backends::{Confirmation, InterfaceIndexConfirmation, PodIp, Targets, Vip};
 use crate::netutils::{if_name_for_routing_ip, if_nametoindex};
-use common::{Backend, BackendKey, BackendList, BACKENDS_ARRAY_CAPACITY};
+use common::{Backend, BackendKey, BackendList, ClientKey, TCPBackend, BACKENDS_ARRAY_CAPACITY};
 
 pub struct BackendService {
     backends_map: Arc<Mutex<HashMap<MapData, BackendKey, BackendList>>>,
     gateway_indexes_map: Arc<Mutex<HashMap<MapData, BackendKey, u16>>>,
+    tcp_conns_map: Arc<Mutex<HashMap<MapData, ClientKey, TCPBackend>>>,
 }
 
 impl BackendService {
     pub fn new(
         backends_map: HashMap<MapData, BackendKey, BackendList>,
         gateway_indexes_map: HashMap<MapData, BackendKey, u16>,
+        tcp_conns_map: HashMap<MapData, ClientKey, TCPBackend>,
     ) -> BackendService {
         BackendService {
             backends_map: Arc::new(Mutex::new(backends_map)),
             gateway_indexes_map: Arc::new(Mutex::new(gateway_indexes_map)),
+            tcp_conns_map: Arc::new(Mutex::new(tcp_conns_map)),
         }
     }
 
@@ -51,6 +54,35 @@ impl BackendService {
         backends_map.remove(&key)?;
         let mut gateway_indexes_map = self.gateway_indexes_map.lock().await;
         gateway_indexes_map.remove(&key)?;
+
+        // Delete all entries in our tcp connection tracking map that this backend
+        // key was related to. This is needed because the TCPRoute might have been
+        // deleted with TCP connection(s) still open, so without the below logic
+        // they'll hang around forever.
+        // Its better to do this rather than maintain a reverse index because the index
+        // would need to be updated with each new connection. With remove being a less
+        // frequently used operation, the performance cost is less visible.
+        let mut tcp_conns_map = self.tcp_conns_map.lock().await;
+        for item in tcp_conns_map
+            .iter()
+            .collect::<Vec<Result<(ClientKey, TCPBackend), MapError>>>()
+        {
+            match item {
+                Ok((
+                    client_key,
+                    TCPBackend {
+                        backend: _,
+                        backend_key,
+                        state: _,
+                    },
+                )) => {
+                    if backend_key == key {
+                        tcp_conns_map.remove(&client_key)?;
+                    };
+                }
+                Err(err) => return Err(err.into()),
+            };
+        }
         Ok(())
     }
 }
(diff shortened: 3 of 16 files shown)


dataplane/ebpf/src/utils.rs
```diff
@@ -34,3 +37,61 @@ pub fn csum_fold_helper(mut csum: u64) -> u16 {
     }
     return !(csum as u16);
 }
+
+// Handles the termination of connection.
+// Ref: https://en.wikipedia.org/wiki/File:Tcp_state_diagram.png and
+// http://www.tcpipguide.com/free/t_TCPConnectionTermination-2.htm
+#[inline(always)]
+pub fn handle_tcp_conn_close(hdr: &TcpHdr, state: &mut TCPState) -> bool {
```
shaneutt, line 45: So while I think the main _reason_ we created this function is to handle closure, in a vacuum what it actually does is something more along the lines of checking if the state needs to be transitioned, right? So similar to my above comment, do you think this naming could be helpful?

```suggestion
pub fn handle_tcp_conn_next_state(hdr: &TcpHdr, state: &mut TCPState) -> bool {
```

Or something like that? Just curious as to your thoughts :thinking:

internal/dataplane/client/utils.go
```diff
@@ -165,13 +160,14 @@ func CompileTCPRouteToDataPlaneBackend(ctx context.Context, c client.Client, tcp
...
-		// TODO(aryan9600): Add support for multiple targets (https://github.com/kubernetes-sigs/blixt/issues/119)
-		Targets: []*Target{target},
+		Targets: backendTargets,
 	}
 
 	return targets, nil
 }
 
+// func makeBackendTargets (ctx context.Context, c client.Client, backendRefs []gatewayv1beta1.BackendRef, ns string) ([]*Target, error)
+
```
shaneutt, line 170: leftover?

```suggestion
```

shaneutt (changes requested):
This looks _great_, thank you for jumping on this!

Many of my comments are about small improvements or idioms but nothing major. I think this is really cool. :tada: 

As an aside and unrelated to any specific bit of code or suggestion: just a general reminder that we're in a memory constrained environment of 512 bytes stack memory with ebpf programs, which felt kind of relevant to me while reading this PR as we've added quite a bit to those programs here. I don't think there's any immediate action to take, as I don't think we're anywhere near in danger of exhausting that limit right now but just something to think about as we continue to add functionality as if we can do _some_ of our optimizations up front we can potentially reduce the size of refactors later.

config/tests/tcproute-rr/patch.yaml
```diff
@@ -0,0 +1,18 @@
...
+  rules:
+  - backendRefs:
+    - name: blixt-tcproute-sample
+      port: 8080
+  - backendRefs:
+    - name: tcproute-rr-v1
+      port: 8080
+  - backendRefs:
+    - name: tcproute-rr-v2
+      port: 8080
```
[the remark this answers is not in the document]
aryan9600: yeah we should be able to, lets do a mix like:
```yaml
  - backendRefs:
    - name: blixt-tcproute-sample
      port: 8080
  - backendRefs:
    - name: tcproute-rr-v1
      port: 8080
    - name: tcproute-rr-v2
      port: 8080
```

config/tests/tcproute-rr/server.yaml
```diff
@@ -0,0 +1,76 @@
...
+      labels:
+        app: tcproute-rr-v1
+    spec:
+      containers:
+      - name: tcp-echo
+        image: istio/tcp-echo-server:1.1
+        imagePullPolicy: IfNotPresent
+        args: [ "8080", "tcproute-rr-v1:" ]
+        ports:
+        - containerPort: 8080
```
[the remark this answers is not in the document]
aryan9600: i'm not quite sure what you are trying to ask. each `backendRef` points to a `Service` which in turn points an `Endpoints` object, so by making sure that requests are distributed in a round-robin fashion, we implicitly test that the control plane is resolving everything properly.

dataplane/ebpf/src/egress/tcp.rs
```diff
@@ -68,6 +71,39 @@ pub fn handle_tcp_egress(ctx: TcContext) -> Result<i32, i64> {
     unsafe { (*tcp_hdr).check = 0 };
 
     // TODO: connection tracking cleanup https://github.com/kubernetes-sigs/blixt/issues/85
```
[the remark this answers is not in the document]
aryan9600: depends on what the scope of the issue is, since its a bit vague. i'll add this back for now, we can revisit it later.

dataplane/ebpf/src/egress/tcp.rs
```diff
@@ -68,6 +71,39 @@ pub fn handle_tcp_egress(ctx: TcContext) -> Result<i32, i64> {
...
+    let mut tcp_state = tcp_backend.state;
+    let moved = handle_tcp_conn_close(tcp_hdr_ref, &mut tcp_state);
+    // If the connection has moved to the Closed state, stop tracking it.
+    if let TCPState::Closed = tcp_state {
+        unsafe {
+            TCP_CONNECTIONS.remove(&client_key)?;
+        }
+    // If the connection has not reached the Closed state yet, but it did advance to a new state,
+    // then record the new state.
+    } else if moved {
```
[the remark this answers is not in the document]
aryan9600: so from what i can understand, we modify `handle_tcp_conn_close` to accept an immutable reference to the backend's state and then return a new state if needed?
this is what i had initially, but decided to change it to the current logic mainly to avoid having to create a new object and reduce memory usage. but then again, this is just an enum so it might be a insignificant optimization. wdyt?

dataplane/ebpf/src/utils.rs
aryan9600: yeah this sounds better (sorry i suck at naming things)

dataplane/ebpf/src/egress/tcp.rs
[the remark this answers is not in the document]
shaneutt: Well this is me suggesting to _remove_ one, not me asking about one that was removed. If there's still more to do it might be worth creating a follow up issue to enumerate those things and update the TODO(s) accordingly?

dataplane/ebpf/src/egress/tcp.rs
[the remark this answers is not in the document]
shaneutt: > so from what i can understand, we modify handle_tcp_conn_close to accept an immutable reference to the backend's state and then return a new state if needed?

Yeah, this is kinda what I was leaning towards? But I wasn't feeling super strongly about it, I was more curious what your thoughts were.

>  this is what i had initially, but decided to change it to the current logic mainly to avoid having to create a new object and reduce memory usage. but then again, this is just an enum so it might be a insignificant optimization. wdyt?

I kinda started down the path of doing a performance analysis here for us to consider together, but quickly realized it would be time consuming and that it shouldn't hold us up from getting something that _works_ first. So let's not worry to much about optimizations at this level just yet, I'm satisfied with what's already here and we can come back around and look these things over in later iterations. :+1:

test/integration/tcproute_test.go
```diff
@@ -69,38 +73,184 @@ func TestTCPRouteBasics(t *testing.T) {
...
+		}
+	}
+	return nil, false
+}
+
+func writeAndReadTCP(t *testing.T, conn net.Conn) string {
+	t.Helper()
+
+	t.Logf("writing data to TCP connection with backend %s", conn.RemoteAddr().String())
+	request := "wazzzaaaa"
```
shaneutt, line 247: lol again

shaneutt (commented):
Alright finally got around to a second pass, no further comments. I'm +1 for it once the open comments are resolved.

also /cc @astoycos for additional review.

dataplane/ebpf/src/egress/tcp.rs
```diff
@@ -68,6 +71,39 @@ pub fn handle_tcp_egress(ctx: TcContext) -> Result<i32, i64> {
...
+        let bk = *tcp_backend;
+        let new_tcp_backend = TCPBackend {
+            backend: bk.backend,
+            backend_key: bk.backend_key,
+            state: tcp_state,
+        };
+        unsafe {
+            TCP_CONNECTIONS.insert(&client_key, &new_tcp_backend, 0_u64)?;
+        }
```
astoycos, line 105: nit: can you just internalize all of this into `handle_tcp_conn_close` or an equalivalent helper function?

dataplane/common/src/lib.rs
```diff
@@ -40,3 +40,39 @@ pub struct BackendList {
...
+    pub ip: u32,
+    pub port: u32,
+}
+
+#[cfg(feature = "user")]
+unsafe impl aya::Pod for ClientKey {}
+
+#[derive(Copy, Clone, Debug, Default)]
+#[repr(C)]
+pub enum TCPState {
```
astoycos, line 56: So this isn't fully descriptive right? We're using map hueristics to determine weather or not the connection is "new" rather than tracking packets through "Start", "Syn_Sent", "Syn_Rcvd" etc right? 

I if ^^ is the case we need some comments as such on this.  Since when I see `TCPState` it makes me think it should contain all possibel states for a TCP connection

dataplane/ebpf/src/ingress/tcp.rs
```diff
@@ -82,15 +131,57 @@ pub fn handle_tcp_ingress(ctx: TcContext) -> Result<i32, i64> {
...
+    // If the packet has the RST flag set, it means the connection is being terminated, so remove it
+    // from our map.
+    if tcp_hdr_ref.rst() == 1 {
+        unsafe {
+            TCP_CONNECTIONS.remove(&client_key)?;
+        }
+    }
 
+    let moved = handle_tcp_conn_next_state(tcp_hdr_ref, &mut tcp_state);
+    // If the connection has moved to the Closed state, stop tracking it.
```
astoycos, line 164: nit: again can we move all of this to utils?

dataplane/common/src/lib.rs
```diff
@@ -40,3 +40,39 @@ pub struct BackendList {
 
 #[cfg(feature = "user")]
 unsafe impl aya::Pod for BackendList {}
+
+#[derive(Copy, Clone, Debug)]
+#[repr(C)]
+pub struct ClientKey {
+    pub ip: u32,
```
astoycos, line 47: I know we don't have an issue for it yet but at some point we'll need to think about v6 here :)

dataplane/api-server/src/server.rs
```diff
@@ -51,6 +54,40 @@ impl BackendService {
         backends_map.remove(&key)?;
         let mut gateway_indexes_map = self.gateway_indexes_map.lock().await;
         gateway_indexes_map.remove(&key)?;
+
+        // Delete all entries in our tcp connection tracking map that this backend
```
astoycos, line 58: Do we need to think about adding a "timeout" at some point so that stale threads are automatically pruned?

astoycos (commented):
Looks really good @aryan9600!!! Mostly nit's and some questions for yah which might result in some follow up issues

Thanks for all the hard work here!

dataplane/loader/src/main.rs
```diff
@@ -46,9 +46,21 @@ async fn main() -> Result<(), anyhow::Error> {
                 .expect("no maps named GATEWAY_INDEXES"),
         )
         .try_into()?;
+        let tcp_conns: HashMap<_, ClientKey, TCPBackend> = Map::HashMap(
```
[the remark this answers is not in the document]
aryan9600: see the PR description :)

dataplane/api-server/src/server.rs
astoycos: Correct, I'm not even sure we need that mechanism yet just thinking about it, Let's defer for a follow up.

astoycos (approved):
Looks great thanks @aryan9600!

Issue #175, opened 2024-01-24
Update README for how to deploy on local


Shunpoco:
Hi communities! :wave: 

I'm very interested in this project, so I tried to deploy Blixt on my local Kind cluster today.
However, I got errors about pulling images from ghcr.io/kubernetes-sigs:
```bash
Warning  Failed     31s (x3 over 76s)  kubelet            Failed to pull image "ghcr.io/kubernetes-sigs/blixt-dataplane:latest": rpc error: code = Unknown desc = failed to pull and unpack image "ghcr.io/kubernetes-sigs/blixt-dataplane:latest": failed to resolve reference "ghcr.io/kubernetes-sigs/blixt-dataplane:latest": failed to authorize: failed to fetch anonymous token: unexpected status from GET request to https://ghcr.io/token?scope=repository%3Akubernetes-sigs%2Fblixt-dataplane%3Apull&service=ghcr.io: 401 Unauthorized
```

After several research, I found a way to solve the problem at https://github.com/kubernetes-sigs/blixt/issues/47.
Now I can deploy Blixt on my cluster by following the steps:
1. `make build.all.images` (as described in https://github.com/kubernetes-sigs/blixt/issues/47#issuecomment-1838654890)
2. `make load.all.images`
3. `kubectl apply -k ./config/tests/integration` (because those built images have a tag `integration-tests`)

I think we should add those steps to README.
I understand that the current situation (using ghcr.io) is temporary, but at the moment, writing the current deployment steps is very user-friendly, especially for newbies.

If those make sense for you, I will make a PR to modify README!

shaneutt:
Thanks @Shunpoco,

Very glad to see your interest in the project, please do reach out to us if there's anything specific you're interested in, if you get stuck and need help, e.t.c.

Good call on the README, it's true that presently the ghcr.io images are not public due to policies about images in `kubernetes-sigs`, and we have an issue (#106) that tracks getting us moved over to an accepted registry. Please do feel free to update the README and add steps to the `# Usage` section, and thanks for taking the time to work through that and identify the problem!

Also note that the tag can be easily overridden, for instance:

```console
$ make build.all.images TAG=latest
$ make load.all.images TAG=latest
```

And then the standard `$ kubectl apply -f ./config/default` should then work.

Shunpoco:
@shaneutt 
Thanks!

I will make a PR soon.

> Also note that the tag can be easily overridden, for instance:

I didn't know that. I will use it for my PR :smile:

PR #177, opened 2024-01-25
Modify README/usage to address the current situation

Shunpoco:
Currently, our image repository (ghcr.io) is private. So users should build and load images themselves.
This PR modifies The usage part in README to add that information.

Fixes: #175

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -331,11 +331,13 @@ build.cluster: $(KTF) # builds a KIND cluster which can be used for testing and 
 .PHONY: load.image
 load.image: build.image
 	kind load docker-image $(BLIXT_CONTROLPLANE_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
-		kubectl -n blixt-system rollout restart deployment blixt-controlplane
+		kubectl -n blixt-system get deployment blixt-controlplane >/dev/null 2>&1 && \
+		kubectl -n blixt-system rollout restart deployment blixt-controlplane || true
 
 .PHONY: load.all.images
 load.all.images: build.all.images
 	kind load docker-image $(BLIXT_CONTROLPLANE_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
 	kind load docker-image $(BLIXT_DATAPLANE_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
 	kind load docker-image $(BLIXT_UDP_SERVER_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
-		kubectl -n blixt-system rollout restart deployment blixt-controlplane
+		kubectl -n blixt-system get deployment blixt-controlplane >/dev/null 2>&1 && \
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -81,13 +81,30 @@ to have _you_ join us in iterating on it and helping us build it together!
...
+```console
+make build.all.images TAG=latest
+```
+
+3. Load images into your Kind cluster:
+
+```console
+make load.all.images TAG=latest
+```
+You may get an error like `Error from server (NotFound): namespaces "blixt-system" not found`, but please ignore it at this moment. The entire resources are created in the next step.
```
shaneutt, line 104: At first I was a bit confused about this, but then taking a look at the `Makefile`, this is because of the rollout restart we added for the load which is convenient for quick building and testing. I had forgotten about that!

If you wanted to as a part of this PR, you could short circuit the error message by doing something like:

```diff
- kubectl -n blixt-system rollout restart deployment blixt-controlplane
+ kubectl -n blixt-system get deployment blixt-controlplane 2>/dev/null && \
+      kubectl -n blixt-system rollout restart deployment blixt-controlplane
```

I just wrote that up here in Github so I didn't test it, but I think something like that should work to only do the rollout if the thing exists, and quell the error output and ignore it if it doesn't.

Shunpoco: Good suggestion! I will modify the Makefile.

Makefile
```diff
@@ -331,11 +331,13 @@ build.cluster: $(KTF) # builds a KIND cluster which can be used for testing and
...
+		kubectl -n blixt-system rollout restart deployment blixt-controlplane || true
 
 .PHONY: load.all.images
 load.all.images: build.all.images
 	kind load docker-image $(BLIXT_CONTROLPLANE_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
 	kind load docker-image $(BLIXT_DATAPLANE_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
 	kind load docker-image $(BLIXT_UDP_SERVER_IMAGE):$(TAG) --name $(KIND_CLUSTER) && \
-		kubectl -n blixt-system rollout restart deployment blixt-controlplane
+		kubectl -n blixt-system get deployment blixt-controlplane >/dev/null 2>&1 && \
+		kubectl -n blixt-system rollout restart deployment blixt-controlplane || true
```
Shunpoco, line 343: 1. I thought that the result of `get deployment` was not needed, so the command sends both stdout and stderr to /dev/null
2. Add `|| true` to finish with status 0 if `blixt-controlplane` is not found

README.md
Shunpoco: Addressed in https://github.com/kubernetes-sigs/blixt/pull/177/commits/938c6866b583745402c5bf704161bfba9e141994

shaneutt (approved):
Awesome, Thanks @Shunpoco!!

/approve
/lgtm

PR #178, opened 2024-01-27
Add support for cross arch docker builds

aryan9600:
Add support for cross architeture Docker builds. Since Arch linux doesn't have an official image for arm64, switch to Ubuntu for building the dataplane.

### DIFF
--- a/dataplane/Dockerfile
+++ b/dataplane/Dockerfile
@@ -1,26 +1,42 @@
-FROM archlinux as builder
+FROM ubuntu:23.10 as builder
 
-RUN pacman -Syu --noconfirm
-RUN pacman -S base-devel protobuf rustup --noconfirm
+ARG TARGETARCH
 
+RUN apt update
+RUN apt install --yes build-essential protobuf-compiler curl pkg-config llvm
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
+
+ENV PATH="/root/.cargo/bin:${PATH}"
 RUN rustup default stable
 RUN rustup install nightly
 RUN rustup component add rust-src --toolchain nightly
-RUN rustup target add x86_64-unknown-linux-musl
 RUN --mount=type=cache,target=/root/.cargo/registry \
     cargo install bpf-linker
 
 WORKDIR /workspace
+# Docker uses the amd64/arm64 convention while Rust uses the x86_64/aarch64 convention.
+# Since Dockerfile doesn't support conditional variables (sigh), write the arch in Rust's
+# convention to a file for later usage.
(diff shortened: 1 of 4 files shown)


dataplane/Dockerfile
```diff
@@ -1,26 +1,42 @@
-FROM archlinux as builder
+FROM ubuntu:23.10 as builder
 
-RUN pacman -Syu --noconfirm
-RUN pacman -S base-devel protobuf rustup --noconfirm
+ARG TARGETARCH
 
+RUN apt update
+RUN apt install --yes build-essential protobuf-compiler curl pkg-config llvm
```
shaneutt, line 6: If I'm not remembering wrong, `apt` is the frontend for a human admin, and `apt-get` is meant for things like automated use cases?

```suggestion
RUN apt-get update
RUN apt-get install --yes build-essential protobuf-compiler curl pkg-config llvm
```

dataplane/Dockerfile
```diff
@@ -1,26 +1,42 @@
...
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
```
shaneutt, line 7: I did like with Arch that `rustup` was one of the signed packages in the default repositories and not just a `curl` into root `bash` line...

Looking back on this change, it seems the goal was to enable cross-compile, but we switched to Ubuntu it would seem to enable building ON non x86 machines... do you think it would be reasonable to stick with Arch as the builder and just enable cross-compilation?

dataplane/Dockerfile
shaneutt, line 6: `apt` (as opposed to `apt-get`) is "meant for end-users" and doesn't guarantee backwards compatibility, so it is generally preferable to use the lower-level `apt-get`:

```suggestion
RUN apt-get update
RUN apt-get install --yes build-essential protobuf-compiler curl pkg-config llvm
```

dataplane/Dockerfile
shaneutt: Talked about this in Slack: we can't simply enable cross-compilation, apparently there is a need to build able to build on ARM systems.

So @aryan9600 I think I have a solution: Perhaps we can just use the [Official Rust Image](https://hub.docker.com/_/rust/). This comes with `rustup` by default and is based on Debian, plus it's multi-arch and available on ARM? Can't remember the reason we used Arch as the builder in the past, but I imagine it had something to do with the protocol buffers support, I think we should see if we can get all that working on the `rust` image, LMKWYT? :thinking:

aryan9600: sounds good to me!

aryan9600:
Closed in favor of #183

Issue #51, opened 2022-12-14
Healthchecks for dataplane API


shaneutt:
## Problem Statement

Our current dataplane has no health checks available to plug into Kubernetes. We should add [readiness and liveness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) so that things work gracefully.

## Acceptance Criteria

- [ ] readiness probes now ensure the dataplane API is accepting traffic before marking the `Pod` as `ready`
- [ ] liveness probles verify whether the dataplane continues to function over time, and restarts when there's problems

alejandroEsc:
:hand: I would be interested in helping out here

shaneutt:
Thank you! Ping us if you run into any trouble

Shunpoco:
Hi @shaneutt ,

Can I tale this issue?

shaneutt:
Sure thing! It's pretty open ended, so please do reach out (here, slack) if you run into any trouble or get stuck. :vulcan_salute: 

/assign @Shunpoco

PR #182, opened 2024-01-30
Dataplane: Add gRPC health check 

Shunpoco:
Currently our dataplane server doesn't have health check, so we can't confirm whether the dataplane daemonset runs properly or not.

This PR introduce health check for gRPC to the dataplane server, and add liveness, readiness, and startup probe for the daemonset.

Fixes: #51

### DIFF
--- a/config/dataplane/dataplane.yaml
+++ b/config/dataplane/dataplane.yaml
@@ -28,3 +28,25 @@ spec:
         - name: RUST_LOG
           value: debug
         imagePullPolicy: IfNotPresent
+        # The gRPC API has a slow startup time, so this probe helps to provide some
+        # grace while starting up to avoid unnecessary kills.
+        #
+        # TODO: When we complete https://github.com/kubernetes-sigs/blixt/issues/173
+        # if we decide that we intend to keep the gRPC API around long term, we should
+        # take some time to see if we can clean up and improve the start time overall.
+        startupProbe:
+          grpc:
+            port: 9874
+          failureThreshold: 30
+          periodSeconds: 10
+        livenessProbe:
+          grpc:
+            port: 9874
+          initialDelaySeconds: 5
+          periodSeconds: 5
+        readinessProbe:
+          grpc:
+            port: 9874
+          initialDelaySeconds: 5
+          periodSeconds: 5
+
--- a/dataplane/api-server/src/lib.rs
+++ b/dataplane/api-server/src/lib.rs
@@ -24,9 +24,12 @@ pub async fn start(
     gateway_indexes_map: HashMap<MapData, BackendKey, u16>,
     tcp_conns_map: HashMap<MapData, ClientKey, LoadBalancerMapping>,
 ) -> Result<(), Error> {
+    let (_, health_service) = tonic_health::server::health_reporter();
+
     let server = server::BackendService::new(backends_map, gateway_indexes_map, tcp_conns_map);
     // TODO: mTLS https://github.com/Kong/blixt/issues/50
     Server::builder()
+        .add_service(health_service)
         .add_service(BackendsServer::new(server))
         .serve(SocketAddrV4::new(addr, port).into())
         .await?;
--- a/test/integration/suite_test.go
+++ b/test/integration/suite_test.go
@@ -36,6 +36,7 @@ import (
 	"sigs.k8s.io/gateway-api/pkg/client/clientset/versioned"
 
 	testutils "github.com/kubernetes-sigs/blixt/internal/test/utils"
+	"github.com/kubernetes-sigs/blixt/pkg/vars"
 )
 
 var (
@@ -181,6 +182,9 @@ func TestMain(m *testing.M) {
 	fmt.Println("INFO: waiting for Blixt component readiness")
 	exitOnErr(testutils.WaitForBlixtReadiness(ctx, env))
 
+	fmt.Println("INFO: waiting for Dataplane readiness")
+	exitOnErr(waitForDataplaneReadiness(ctx, env))
+
 	exit := m.Run()
 
 	exitOnErr(runCleanup(mainCleanupKey))
@@ -306,6 +310,53 @@ func waitForBpfdConfigDelete(ctx context.Context, env environments.Environment) 
 				}
 				return err
 			}
+		}
+	}
+}
+
+func waitForDataplaneReadiness(ctx context.Context, env environments.Environment) error {
+	for {
+		select {
+		case <-ctx.Done():
+			if err := ctx.Err(); err != nil {
+				return fmt.Errorf("context completed while waiting for dataplane readiness, and an error occurred: %w", err)
+			}
+			return fmt.Errorf("context completed while waiting for dataplane readiness")
+		default:
+			dataplanes, err := env.Cluster().Client().CoreV1().Pods(vars.DefaultNamespace).
+				List(ctx, metav1.ListOptions{LabelSelector: fmt.Sprintf("component=%s", vars.DefaultDataPlaneComponentLabel)})
+			if err != nil {
+				return fmt.Errorf("failed to fetch dataplane pod list: %w", err)
+			}
+
+			if len(dataplanes.Items) == 0 {
+				return fmt.Errorf("no dataplanes found in namespace %s", vars.DefaultNamespace)
+			}
+
+			// Check whether dataplane is set readiness probe.
+			for _, container := range dataplanes.Items[0].Spec.Containers {
+				if container.Name == "dataplane" && container.ReadinessProbe == nil {
+					return fmt.Errorf("found a dataplane container which doesn't have readiness probe")
+				}
+			}
+
+			// Check all pods' readiness
+			var ready int
+			for _, pod := range dataplanes.Items {
+				for _, status := range pod.Status.ContainerStatuses {
+					if status.Name == "dataplane" && status.Ready {
+						ready++
+					}
+				}
+			}
+
+			if ready != len(dataplanes.Items) {
+				fmt.Printf("%d dataplanes not yet ready\n", len(dataplanes.Items)-ready)
+				time.Sleep(time.Second) // small rest from hitting the API over and over again
+				break
+			}
+
+			return nil
 		}
 	}
 }
--- a/dataplane/api-server/Cargo.toml
+++ b/dataplane/api-server/Cargo.toml
@@ -7,6 +7,7 @@ publish = false
 [dependencies]
 prost = "0.11.9"
 tonic = "0.9.2"
+tonic-health = "0.9.2"
 anyhow = "1"
 log = "0.4"
 aya = { version = ">=0.11", features=["async_tokio"] }


config/dataplane/dataplane.yaml
```diff
@@ -28,3 +28,19 @@ spec:
         - name: RUST_LOG
           value: debug
         imagePullPolicy: IfNotPresent
+        startupProbe:
```
Shunpoco, line 31: Note: I found that sometimes a delay of launching a pod causes liveness/readiness failures without startupProbe.
Of course after several seconds, a pod works properly, but I think setting this startupProbe is appropriate.

dataplane/api-server/src/lib.rs
```diff
@@ -24,9 +24,15 @@ pub async fn start(
     gateway_indexes_map: HashMap<MapData, BackendKey, u16>,
     tcp_conns_map: HashMap<MapData, ClientKey, LoadBalancerMapping>,
 ) -> Result<(), Error> {
+    let (mut health_reporter, health_service) = tonic_health::server::health_reporter();
+    health_reporter
+        .set_serving::<BackendsServer<server::BackendService>>()
+        .await;
```
aryan9600, line 30: don't think this is required, since we are not specifying the service name in any of the probes anyway.

Shunpoco:
@shaneutt 
Let me ask about the verification for readiness of the dataplane container.
`WaitForBlixtReadiness` already checks if the dataplane pod is ready at least in one node [here](https://github.com/kubernetes-sigs/blixt/blob/main/internal/test/utils/utils.go#L126-L133).
I think before adding the readiness probe, k8s decided that the dataplane pod is ready if a container in the pod is not down.
After adding the readiness probe, I think k8s decides that the dataplane pod is ready using the probe.
So the correctness of the result of `WaitForBlixtReadiness` will be improved by the readiness probe, but I'm wondering we should add more verification or current implementation is enough.

shaneutt:
> @shaneutt Let me ask about the verification for readiness of the dataplane container. `WaitForBlixtReadiness` already checks if the dataplane pod is ready at least in one node [here](https://github.com/kubernetes-sigs/blixt/blob/main/internal/test/utils/utils.go#L126-L133). I think before adding the readiness probe, k8s decided that the dataplane pod is ready if a container in the pod is not down (it ignores whether a container shows error or not). After adding the readiness probe, k8s decides if the dataplane pod is ready using the probe. So the correctness of the result of `WaitForBlixtReadiness` will be improved by the readiness probe, but I'm wondering we should add more verification or current implementation is enough.

To me this seems like enough of an iterative improvement for now that we can move forward with this, see how it works out, and maybe down the road considering if we want to do more.

dataplane/api-server/src/lib.rs
Shunpoco: Modified in https://github.com/kubernetes-sigs/blixt/pull/182/commits/82c634e3ca82b376d9266ce8c68d48e01c530e3c

shaneutt:
Hey @Shunpoco just checking in, everything going OK? Need any assistance?

Shunpoco:
@shaneutt 
Thanks for the ping!
I think once @aryan9600 reviews [this change](https://github.com/kubernetes-sigs/blixt/pull/182#discussion_r1473618601) and approves it, this PR is ready for merge :smile:

Shunpoco:
@shaneutt 
OK, let me clarify, I think I misunderstood.
In [my comment](https://github.com/kubernetes-sigs/blixt/pull/182#issuecomment-1920095633), I wanted to say that `WaitForBlixtReadiness` already checks whether pod is ready, so we don't need to put `requires.Eventually()` for the readiness probe.
But do you want me to add an additional codes for check the pod's status?

shaneutt:
> @shaneutt OK, let me clarify, I think I misunderstood. In [my comment](https://github.com/kubernetes-sigs/blixt/pull/182#issuecomment-1920095633), I wanted to say that `WaitForBlixtReadiness` already checks whether pod is ready, so we don't need to put `requires.Eventually()` for the readiness probe. But do you want me to add an additional codes for check the pod's status?

Yeah, my thinking was this: you're right that the current tests effectively exercise the health checks, but nothing is verifying their presence or configuration. A simple couple line check could help us to avoid regressions (though we would hope they would never happen!) or mistakes in test configuration later on and just double check that the health checks are present, and are configured as intended. Let me know what you think?

Shunpoco:
@shaneutt 

I got it. I will add the verification.

I pushed https://github.com/kubernetes-sigs/blixt/pull/182/commits/da1be5fe15229a98a80d2002efe679440159ed0f as a draft of the verification (I will blush up after I confirm we have the same view).

- I put `waitForDataplaneReadiness` function in suite_test.go. The function does: 1) check if the dataplane pod has readinessprobe (if not, return an error), and then 2) check if the all dataplane pods are ready.
- Call the function in TestMain after the dataplane is deployed.

Note: Because TestMain doesn't take testing.T, I don't use require.Eventually.

Does this change suit for your intention?

shaneutt:
Yep, this looks good. Might seem a little superfluous but I appreciate you adding it as a documentation of intent, and on the off chance the health checks fail, now we can fail at a very specific place with specific logging which will be easier for debugging.

I'll mark this draft temporarily since you said you wanted to do a couple more things, please mark as ready and call me in for a final review whenever you're ready. Very much appreciate your contribution!

Shunpoco:
@shaneutt 
Now this PR is ready for review! please check https://github.com/kubernetes-sigs/blixt/pull/182/commits/c7fc30287fc154036daa96988eecbef6c3a437bc.

test/integration/suite_test.go
```diff
@@ -309,3 +313,47 @@ func waitForBpfdConfigDelete(ctx context.Context, env environments.Environment)
...
+		time.Sleep(2 * time.Second)
+		select {
+		case <-ctx.Done():
+			if err := ctx.Err(); err != nil {
+				return fmt.Errorf("context completed while waiting for dataplane ready: %w", err)
+			}
+			return fmt.Errorf("context completed while waiting for dataplane ready")
+		default:
+			fmt.Println("INFO: waiting for dataplane ready")
+
```
shaneutt, line 328: Right before we call this function we already print an info line that essentially matches this, so as written we would end up with:

```console
INFO: waiting for Dataplane readiness
INFO: waiting for dataplane ready
```

and then depending on how long it takes, maybe:

```
INFO: waiting for Dataplane readiness
INFO: waiting for dataplane ready
INFO: waiting for dataplane ready
INFO: waiting for dataplane ready
```

I suggest we remove this line here, and instead emit additional and specific info as necessary further down:

```suggestion
```

test/integration/suite_test.go
```diff
@@ -309,3 +313,47 @@ func waitForBpfdConfigDelete(ctx context.Context, env environments.Environment)
...
+			dataplanes, err := env.Cluster().Client().CoreV1().Pods(vars.DefaultNamespace).
+				List(ctx, metav1.ListOptions{LabelSelector: fmt.Sprintf("component=%s", vars.DefaultDataPlaneComponentLabel)})
+			if err != nil {
+				return fmt.Errorf("failed to fetch dataplane pod list: %w", err)
+			}
+
+			// Check whether dataplane is set readiness probe.
+			for _, container := range dataplanes.Items[0].Spec.Containers {
```
shaneutt, line 336: While we would really hope that listing the dataplanes at this point would not return an empty list, it technically could and then we would get an `index out of range` panic here. We throw an error if the listing fails, we could also throw an error if the list succeeds but is unexpectedly empty:

```suggestion
			if len(dataplanes.Items) == 0 {
				return fmt.Errorf("no dataplanes found in namespace %s", vars.DefaultNamespace)
			}

			// Check whether dataplane is set readiness probe.
			for _, container := range dataplanes.Items[0].Spec.Containers {
```

test/integration/suite_test.go
```diff
@@ -309,3 +313,47 @@ func waitForBpfdConfigDelete(ctx context.Context, env environments.Environment)
...
+				if container.Name == "dataplane" && container.ReadinessProbe == nil {
+					return fmt.Errorf("container doesn't have readiness probe")
```
shaneutt, line 338:
```suggestion
					return fmt.Errorf("found a dataplane container which doesn't have readiness probe")
```

test/integration/suite_test.go
```diff
@@ -309,3 +313,47 @@ func waitForBpfdConfigDelete(ctx context.Context, env environments.Environment)
...
+			for _, pod := range dataplanes.Items {
+				for _, status := range pod.Status.ContainerStatuses {
+					if status.Name == "dataplane" && status.Ready {
+						ready++
+					}
+				}
+			}
+
+			if ready != len(dataplanes.Items) {
+				return fmt.Errorf("dataplane not ready")
```
shaneutt, line 353: We have two returns above with errors that indicate problems we don't expect to have while waiting for the dataplane (API connectivity issues and missing readiness probes) which are both good checks.

However _this_ return creates an early exit and basically means that the `for` loop which encompasses it is not practically exercised, we could break here instead of returning an error:

```suggestion
				fmt.Printf("%d dataplanes not yet ready", len(dataplanes.Items) - ready)
				time.Sleep(time.Second) // small rest from hitting the API over and over again
				break
```

This will make sure the `for` loop takes effect and retries the listing when the pods are not yet ready. Then too we can put the `time.Sleep` here so that we only pause if needed to between API calls.

shaneutt (changes requested):
Looking good!

Couple minor comments to sort out then I think we're ready to :ship:

test/integration/suite_test.go
Shunpoco: Nice, I misunderstood exitonError's behaviour

test/integration/suite_test.go
```diff
@@ -309,3 +313,47 @@ func waitForBpfdConfigDelete(ctx context.Context, env environments.Environment)
 		}
 	}
 }
+
+func waitForDataplaneReadiness(ctx context.Context, env environments.Environment) error {
+	for {
+		time.Sleep(2 * time.Second)
```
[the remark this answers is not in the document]
Shunpoco: This is very arbitrary. I will remove it :)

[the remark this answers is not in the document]
Shunpoco: delete at https://github.com/kubernetes-sigs/blixt/pull/182/commits/d7951ec1c6749f22986c50a3c37c11a5518d2a66

test/integration/suite_test.go
```diff
@@ -309,3 +313,50 @@ func waitForBpfdConfigDelete(ctx context.Context, env environments.Environment)
 		}
 	}
 }
+
+func waitForDataplaneReadiness(ctx context.Context, env environments.Environment) error {
+	for {
+		select {
+		case <-ctx.Done():
+			if err := ctx.Err(); err != nil {
+				return fmt.Errorf("context completed while waiting for dataplane ready, and an error occurred: %w", err)
```
aryan9600, line 322:
```suggestion
				return fmt.Errorf("context completed while waiting for dataplane readiness, and an error occurred: %w", err)
```

aryan9600 (commented):
looks pretty good, just have a couple of nits. also it'd be great if you could squash all your commits into 1-2 concise commits, thanks!

aryan9600 (approved):
LGTM!
Thanks @Shunpoco 🎖️

shaneutt (approved):
/lgtm

PR #184, opened 2024-02-11
Add a new gateway reconciler in rust

### LINKED ISSUE #176: Port the Gateway reconciler to Rust
With https://github.com/kubernetes-sigs/blixt/discussions/150, we decided to rewrite the controlplane in Rust. The best way to go about it would be to do it in phases since there exist multiple reconcilers. Beginning with the Gateway reconciler makes the most sense since the other two reconcilers are similar and thus could be tackled at the same time. The main goal here is to make sure we do a safe port and don't break any existing functionality, barring the odd low hanging fruit.

aryan9600:
Introduce a new folder `controlplane-rs` meant for the new controlplane that's to be written in Rust (https://github.com/kubernetes-sigs/blixt/discussions/150). Add a new reconciler for Gateway objects that behaves pretty similarly to the reconciler written in Golang (along with a few improvements around conditions and reasons). It reconciles Gateway v1 objects unlike the Go reconciler which works with v1beta1 objects.

TODOs:
- [ ] tests
- [ ] utilities to build and deploy
- [ ] a convenient way to swap out the Golang reconciler for this one

Fixes #176


astoycos:
YAYYYY I'll try and get to this during the week!! Thanks @aryan9600!!!!

shaneutt:
@aryan9600 you can rebase to the latest `main` now for some general CI fixes that just merged.

controlplane-rs/.dockerignore
```diff
@@ -0,0 +1,3 @@
+.vim/
+target/
+*.md
```
shaneutt, line 3: I think longterm it would make sense to just do `controlplane` as opposed to `controlplane-rs`, right? We need to think about how this directory is going to look to someone new coming to the codebase who might not immediately understand the scope of our transition. A README in this directory explaining the transition and mentioning the Go code is the current controlplane for the time being would be helpful.

controlplane-rs/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM --platform=$BUILDPLATFORM tonistiigi/xx AS xx
+
```
shaneutt, line 2: What is the `xx` image all about? I'm not sure about this :thinking:

controlplane-rs/src/gateway_controller.rs
```diff
@@ -0,0 +1,178 @@
...
+use chrono::Utc;
+use gateway_utils::*;
+use k8s_openapi::apimachinery::pkg::apis::meta::v1 as metav1;
+use tracing::*;
+
+pub async fn reconcile(gateway: Arc<Gateway>, ctx: Arc<Context>) -> Result<Action> {
+    let start = Instant::now();
+    let client = ctx.client.clone();
+    let name = gateway.name_any();
+    let ns = gateway.namespace().unwrap();
```
shaneutt, line 42: A flat `unwrap()` here might be problematic, if I'm not mistaken you can in fact have no namespace declared and that will mean the resource belongs to the default namespace right? If you find that's accurate, then we can probably do something like:

```suggestion
    let ns = gateway.namespace().unwrap_or("default".into());
```

Which would be safe. However we should ideally use an existing constant for `"default"` if one exists in kube-rs or somewhere nearby.
```

**Review by shaneutt — COMMENTED** [review]
Got an initial start on a PR, but didn't get super far yet, I will come back around!

File: controlplane-rs/src/gateway_controller.rs (hunk 3 above)
Comment (shaneutt) [inline controlplane-rs/src/gateway_controller.rs:42]: A flat `unwrap()` is problematic, it seems like we could probably do:

```suggestion
    let ns = gateway.namespace().unwrap_or("default".into());
```

Which would be safe. However we should ideally use an existing constant for `"default"` if one exists in kube-rs or somewhere nearby.

controlplane-rs/src/gateway_controller.rs
```diff
@@ -0,0 +1,178 @@
...
+        status: gateway.status.clone(),
+    };
+
+    let gateway_class_api = Api::<GatewayClass>::all(client.clone());
+    let gateway_class = gateway_class_api
+        .get(gateway_spec.gateway_class_name.as_str())
+        .await
+        .map_err(Error::KubeError)?;
+
+    let gateway_class_spec: GatewayClassSpec = gateway_class.spec.clone();
```
shaneutt, line 58: This stuff looks like setup that we might be able to re-use some between all the controllers, what do you think about creating some utility functions for this?

controlplane-rs/src/gateway_controller.rs
```diff
@@ -0,0 +1,178 @@
...
+    if gateway_class_spec.controller_name.as_str() != GATEWAY_CLASS_CONTROLLER_NAME {
+        return Ok(Action::requeue(Duration::from_secs(3600)));
```
shaneutt, line 60: If the `controller_name` doesn't match, we don't support this resource so we shouldn't re-queue we should [possibly debug log and then] drop right?

controlplane-rs/src/gateway_controller.rs
```diff
@@ -0,0 +1,178 @@
...
+    let accepted_cond = get_accepted_condition(&mut gw);
+    set_condition(&mut gw, accepted_cond.clone());
+    if accepted_cond.status == "False".to_string() {
+        let programmed_cond = metav1::Condition {
+            last_transition_time: accepted_cond.last_transition_time.clone(),
+            observed_generation: accepted_cond.observed_generation,
+            type_: "Programmed".to_string(),
+            status: "False".to_string(),
+            message: accepted_cond.message.clone(),
+            reason: "Invalid".to_string(),
```
shaneutt, line 77: Ideally we should be getting these from the Gateway API, similar to what can be done using the Golang library for Go controllers. We can put in PRs to https://github.com/kube-rs/gateway-api-rs as needed.

controlplane-rs/src/gateway_controller.rs
```diff
@@ -0,0 +1,178 @@
...
+        };
+        set_condition(&mut gw, programmed_cond);
+    }
+
+    if accepted_cond.status == String::from("False") {
```
shaneutt, line 82: You don't have to explicitly coerce a `&str` into a `String` to compare it to another `String` in Rust, you can instead just use the comparison operator directly between the two types:

```rust
if accepted_cond.status == "False" {
```

and under the hood Rust will provide automatic dereferencing of the `String` to a `&str` so that they can be compared.

Also, can we get the condition statuses from `k8s_openapi` instead? :thinking:

controlplane-rs/src/gateway_controller.rs
```diff
@@ -0,0 +1,178 @@
...
+        patch_status(&gateway_api, name, gw.status.as_ref().unwrap()).await?;
```
shaneutt, line 83: We need to handle rather than `unwrap()`

controlplane-rs/src/gateway_controller.rs
```diff
@@ -0,0 +1,178 @@
...
+        return Err(Error::InvalidConfigError(accepted_cond.message));
+    }
+
+    let service_api: Api<Service> = Api::namespaced(client, &ns);
+    let services = service_api
+        .list(&ListParams::default().labels(&format!("{}={}", GATEWAY_SERVICE_LAEBL, name)))
```
shaneutt, line 89:
```suggestion
        .list(&ListParams::default().labels(&format!("{}={}", GATEWAY_SERVICE_LABEL, name)))
```

controlplane-rs/src/gateway_controller.rs
```diff
@@ -0,0 +1,178 @@
...
+        .await
+        .map_err(Error::KubeError)?;
+    if services.items.len() > 1 {
+        return Err(Error::LoadBalancerError(
+            "found more than 1 Service for this Gateway; multiple services are not supported"
+                .to_string(),
+        ));
+    }
```
shaneutt, line 97: I think for the first pass it's probably OK to leave this here, but we should ideally put a `TODO` comment on it with a link to an issue where we will follow up later, and then we should probably log the names of all the `Service` resources we found in case this ever _does actually happen_ the user has a hint about which `Service` resources are offending.

controlplane-rs/src/gateway_controller.rs
```diff
@@ -0,0 +1,178 @@
...
+                )
+                .await
+                .map_err(Error::KubeError)?;
+        }
+    } else {
+        info!("creating loadbalancer service");
+        service = create_svc_for_gateway(ctx.clone(), gateway.as_ref()).await?;
+    }
+
+    let now = k8s_openapi::apimachinery::pkg::apis::meta::v1::Time(Utc::now());
```
shaneutt, line 120: We should alias

```suggestion
    let now = metav1::Time(Utc::now());
```

controlplane-rs/src/gateway_controller.rs
```diff
@@ -0,0 +1,178 @@
...
+        let msg = "LoadBalancer does not have a ingress IP address".to_string();
+        set_condition(
+            &mut gw,
+            metav1::Condition {
+                last_transition_time: now,
+                observed_generation: gateway.meta().generation,
+                message: msg.clone(),
+                reason: "AddressNotAssigned".to_string(),
+                status: "False".to_string(),
+                type_: "Programmed".to_string(),
```
shaneutt, line 131: Similar to above with getting our conditions from the `gateway-api` crate if we can.

controlplane-rs/src/gateway_controller.rs
```diff
@@ -0,0 +1,178 @@
...
+            },
+        );
+        patch_status(&gateway_api, name, &gw.status.unwrap()).await?;
```
shaneutt, line 134: We need to handle the `unwrap()`.

controlplane-rs/src/gateway_controller.rs
```diff
@@ -0,0 +1,178 @@
...
+    create_endpoint_if_not_exists(ctx.clone(), &service).await?;
+    set_gateway_status_addresses(&mut gw, &service);
+
+    let programmed_cond = metav1::Condition {
+        last_transition_time: now,
+        observed_generation: gateway.meta().generation,
+        type_: "Programmed".to_string(),
+        status: "True".to_string(),
+        reason: "Programmed".to_string(),
+        message: "Dataplane configured for gateway".to_string(),
```
shaneutt, line 147: Similar to above with getting our conditions from the `gateway-api` crate if we can.

controlplane-rs/src/gateway_controller.rs
```diff
@@ -0,0 +1,178 @@
...
+    };
+    set_condition(&mut gw, programmed_cond);
+
+    patch_status(&gateway_api, name, gw.status.as_ref().unwrap()).await?;
```
shaneutt, line 151: Need to handle the `unwrap()`s

shaneutt (commented):
Took another review pass today, but I still haven't gotten entirely through it. Mostly minor things so far, looking good! I will take another pass soon.

astoycos (commented):
Yahoooo Let's get this rolling :) 

I still have way more to go here but I probably won't be able to get to it until tonight/tomorrow 

it's just a reconciler for the gateway object not gatewayclass, tcproute, or udproute at the moment right?

shaneutt:
@aryan9600 looks like we need some updates on your side here to push on, let us know if you're stuck or blocked or need any support :vulcan_salute:

aryan9600:
@shaneutt @astoycos this is ready for another round of review :)

controlplane-rs/src/gateway_controller.rs
aryan9600: is there any scenario where we would actually want to support multiple `Service`s? If a `Gateway` has N listeners, we can just define N ports on the `Service` itself (which is what we do atm).

controlplane-rs/src/gateway_controller.rs
aryan9600: not exactly sure how this can be re-used between controllers 😅 
everything here is statically typed, so can't think how we can have generic util functions for this

controlplane-rs/Dockerfile
aryan9600: https://github.com/tonistiigi/xx
tldr: it contains helper scripts for common tools like `cargo`, `go`, etc. to make building cross platform docker images easy

shaneutt: These kinds of dependencies always make me a bit wary. If you think it adds a lot of value then maybe we can move forward with it, but if its only keeping us from writing say, a couple lines? I would consider removing it from the supply-chain.

controlplane-rs/src/gateway_controller.rs
shaneutt: I'd have to think about it. I guess for now so we can move forward quickly I think it would be reasonable to just add a TODO here to re-evaluate, and answer that question.

shaneutt (approved):
Thanks @aryan9600 this is great!

While I'm sure we _could_ do several more rounds of review on this, given the sandbox nature of the project and the foundational nature of this work (we have others waiting to contribute to the new Rust control-plane) I'm in favor of defaulting to shipping this for now so we can move quickly into the other parallel iterations that will use this as groundwork.

/approve

@astoycos please lgtm when you're ready

controlplane-rs/Dockerfile
EandrewJones: I second Shane's concern. Looking through the repo, I can see the value-add---setting up cross-platform images can be a pain in the booty.

That said, I'm always inclined to err on the side of keeping your dependency footprint smaller, especially for security purposes.

shaneutt: What are your thoughts on this one @aryan9600?

shaneutt: Since this is sorta blocking all other work on the control-plane, I opted for a follow-up so we don't have to hold this PR up: https://github.com/kubernetes-sigs/blixt/issues/256

PR #197, opened 2024-02-20
Fix: Correct TCP connection tracking state update

tonsiasy:
I fixed a bug related to TCP state updates.


shaneutt (approved):
Thank you @tonsiasy, very good catch!

For posterity (and since I know Rust is fairly new to some Kubernetes SIGs members and there's several people out there watching and learning), it might be helpful to provide an explanation of how this fix works:

The `TCPState` enum has the `Copy` trait implemented for it, so during the [destructuring](https://doc.rust-lang.org/rust-by-example/flow_control/match/destructuring.html) in `if let Some(mut tcp_state)` we were _copying_ the value of `lb_mapping.tcp_state` which was not correct because of our intention to then mutate that value with the `process_tcp_state_transition()` function (the mutation was only occurring on the _copy_ not the original value).

The change to `if let Some(ref mut tcp_state)` avoids copying in the destructuring and provides us a direct reference to the field within the existing struct so that we can mutate it properly.

For any Rust newcomers, I've created a playground program which should help to illustrate:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=111aee70fefbb4120aa557273e662527

Again thank you @tonsiasy we appreciate you jumping in to help fix this. :partying_face:

PR #211, opened 2024-03-20
Fix builds, bump to aya 0.12.0 and some cleanups

astoycos:
Stop using aya:main and use the recent 0.12.0
release.

Bump some other crate deps

other small fixups too make this all work.

run cargo fmt


aryan9600 (approved):
lgtm! thanks @astoycos 🙇

PR #217, opened 2024-03-29
#212 add cargo format and clippy checks to ci

### LINKED ISSUE #212: Add Cargo format and clippy checks to our CI for all rust code
Similar to what we do in bpfman at -> https://github.com/bpfman/bpfman/blob/main/.github/workflows/build.yml#L84

we should be running `cargo fmt` and `cargo clippy` against all our rust based code to make sure it's formatted and checked correctly.

EandrewJones:
Closes #212.

### Motivation
The code takes a cue from [bpfman](https://github.com/bpfman/bpfman/blob/main/.github/workflows/build.yml#L84) GH action to add clippy and fmt checks in our ci build steps.

The make commands I wrote search for `Cargo.toml` as targets to run clippy and format against. They should automatically extend to new rust components if we add them to the lib.

### Notes
1) I did have to make some changes to code in `dataplane/ebf` to make clippy happy. I followed the recommended changes verbatim, so please take a look at these to make sure I didn't blindly follow clippy's advice and do something stupid.
2) I couldn't get nektos/act to play nice with rust so I didn't test ci.yaml locally. I may have to use this PR as my test environment. :(


EandrewJones:
FYSA: I might need to move the rust ci steps into their own action given that Docker build skips by default and that's where they're currently located.

.github/workflows/ci.yaml
```diff
@@ -6,7 +6,7 @@ on:
   pull_request: {}
   push:
     branches:
-    - main
+      - main
```
astoycos, line 9: what's up with all these added spaces in the diff?

astoycos:
> FYSA: I might need to move the rust ci steps into their own action given that Docker build skips by default and that's where they're currently located.

Feel free to, in bpfman we have separate `build-go` and `build-rust` flows.

astoycos:
I also don't think it'd be the worst thing to have a deidcated `build` workflow that happens irerespective of the container image build pipeline

.github/workflows/ci.yaml
EandrewJones: My B. I didn't notice these, must be VS code triggering some yaml auto-formatting. I'll fix it up.

EandrewJones:
> > FYSA: I might need to move the rust ci steps into their own action given that Docker build skips by default and that's where they're currently located.
> 
> Feel free to, in bpfman we have separate `build-go` and `build-rust` flows.

Makes sense.

>I also don't think it'd be the worst thing to have a dedicated build workflow that happens irrespective of the container image build pipeline that runs on every PR so that we can catch build errors early

Agreed --- I'll take a stab at that and more these rust checks into the rust build flow.

EandrewJones:
@astoycos

I'm running into an annoying tautology with the linting and build CI action.

## Problem
Clippy fails if you do not build the `dataplane` first because it's missing a dependency from `cargo xtask build-ebpf`. Clippy _also_ fails if you run `cargo xtask build-ebpf` first.

Why? As a side effect, it appears to auto-generate/modify formatting of `dataplane/api-server/src/backend.rs` in a way that makes clippy throw.

## Solution
I see 2 possible fixes:
1.  Run `make fix.format.rust` between the build and clippy steps.
2. Get rid of the `cargo xtask build-ebpf` side-effects.

The former defeats the purpose of a check in the first place since it would mask any other clippy violations introduced by the PR as well. The latter is the correct approach but bit beyond my current understanding of the codebase.

Thoughts?

EandrewJones:
Pinging @astoycos @shaneutt @aryan9600. Would someone mind triggering the workflows?

EandrewJones:
> Pinging @astoycos @shaneutt @aryan9600. Would someone mind triggering the workflows?

Thanks Shane.

Hmm...integration tests now fail. It seems like maybe the controller isn't getting a response from the dataplane. My code changes shouldn't have impacted functionality. Will test locally.

Would a more experienced rustacean mind browsing the few changes I made. Maybe there's a red flag I'm unaware of. Thanks!

astoycos:
Looks like the failure was a flake (passed after re-run)

EandrewJones:
Cool.
Best

Evan Jones
Website: www.ea-jones.com

astoycos (commented):
Just one small thing and I need to checkup on that one bit later today thanks so much!

.github/workflows/build.yaml
```diff
@@ -0,0 +1,83 @@
...
+        make build
+    
+    # Clippy fails if you do not build the dataplane first because it's missing a dependency 
+    # from cargo xtask build-ebpf. Clippy also fails if you run cargo xtask build-ebpf first.
+    #
+    # Why? As a side effect, it auto-generates dataplane/api-server/src/backend.rs with formatting
+    # that makes clippy throw. There's no easy way to fix this side-effect. Therefore, we need to 
+    # go in and surgically auto-format the file before we try to run clippy.
+    # This is super ugly, but it works barring a more formal solution. 
+    - name: Fix backends.rs formatting
```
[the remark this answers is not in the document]
astoycos: @EandrewJones Do you mind cherry-picking https://github.com/kubernetes-sigs/blixt/commit/84e2a14de3f00eb482812564a436cfb0e4f58729 and then removing this bit? 

We shouldn't be rebuilding our rust bindings on every build.... since really the api-server crate is just a lib it doesn't/shouldn't be executing anything unless the proto file changes

If cherry-picking is annoying I can open a PR just lmk!

EandrewJones:
cherry picking is fine. Thanks for implementing the clean solution :) I
hate ugly hacks.

Best

Evan Jones
Website: www.ea-jones.com

.github/workflows/build.yaml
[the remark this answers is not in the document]
EandrewJones: Where does this commit live? It doesn't seem to be on origin? Is it on your fork? BLUF: I can't fetch it from anywhere to cherry pick it.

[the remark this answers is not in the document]
astoycos: Soooory it's here -> https://github.com/astoycos/blixt/tree/pr/EandrewJones/217

EandrewJones:
All issues addressed.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,66 @@
...
+  rust-build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -qy \
```
astoycos, line 24: We shouldn't need this anymore

.github/workflows/build.yaml
```diff
@@ -0,0 +1,66 @@
...
+        override: false
+
+    - name: Install bpf-linker
+      run: |
+        cargo install bpf-linker
+
+    # TODO: Once we migrate the controller from Go to Rust,
+    # add the controller build step here.
+
+    - name: Clean UDP Test Server
```
astoycos, line 45: Again do why do we need this clean?

astoycos (commented):
Two small nits, then please squash your commits and this should be good to go 👍 

Thanks @EandrewJones!

.github/workflows/build.yaml
EandrewJones: Oops. Missed this one.

EandrewJones:
> /lgtm /approve

Sorry. The bot complained about my commit message so I revised it to remove the problematic keywords.

astoycos:
No worries 

/lgtm
/approve

PR #220, opened 2024-04-07
refactor: optimize TCP and UDP checksum handling with eBPF

tzssangglass:
udp and tcp ingress:
1. use `bpf_l4_csum_replace` `bpf_l3_csum_replace` to recalculate the checksums;
2. and `bpf_skb_store_bytes` to update dest ip and port;

Signed-off-by: tzssangglass <tzssangglass@gmail.com>

### DIFF
--- a/dataplane/ebpf/src/ingress/tcp.rs
+++ b/dataplane/ebpf/src/ingress/tcp.rs
@@ -6,30 +6,21 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
 
 use core::mem;
 
-use aya_ebpf::{
-    bindings::TC_ACT_OK,
-    helpers::{bpf_l4_csum_replace, bpf_l3_csum_replace, bpf_skb_store_bytes, bpf_redirect_neigh},
-    programs::TcContext,
-};
+use aya_ebpf::{bindings::TC_ACT_OK, helpers::bpf_redirect_neigh, programs::TcContext};
 use aya_log_ebpf::{debug, info};
-use aya_ebpf_cty::c_void;
 
-use network_types::{eth::EthHdr, ip::Ipv4Hdr, tcp::TcpHdr};
 use memoffset::offset_of;
+use network_types::{eth::EthHdr, ip::Ipv4Hdr, tcp::TcpHdr};
 
 use crate::{
-    utils::{ptr_at, update_tcp_conns},
+    utils::{ptr_at, set_ipv4_dest_port, set_ipv4_ip_dst, update_tcp_conns},
     BACKENDS, GATEWAY_INDEXES, LB_CONNECTIONS,
 };
 use common::{
     Backend, BackendKey, ClientKey, LoadBalancerMapping, TCPState, BACKENDS_ARRAY_CAPACITY,
 };
 
-const IP_CSUM_OFF: u32 = (EthHdr::LEN + offset_of!(Ipv4Hdr, check)) as u32;
 const TCP_CSUM_OFF: u32 = (EthHdr::LEN + Ipv4Hdr::LEN + offset_of!(TcpHdr, check)) as u32;
-const IP_DST_OFF: u32 = (EthHdr::LEN + offset_of!(Ipv4Hdr, dst_addr)) as u32;
-const TCP_DPORT_OFF: u32 = (EthHdr::LEN + Ipv4Hdr::LEN + offset_of!(TcpHdr, dest)) as u32; 
-const IS_PSEUDO: u64 = 0x10;
 
 pub fn handle_tcp_ingress(ctx: TcContext) -> Result<i32, i64> {
     let ip_hdr: *mut Ipv4Hdr = unsafe { ptr_at(&ctx, EthHdr::LEN)? };
@@ -114,23 +105,10 @@ pub fn handle_tcp_ingress(ctx: TcContext) -> Result<i32, i64> {
         u16::from_be(unsafe { (*tcp_hdr).dest })
     );
 
-    // DNAT the ip address
-    unsafe {
-        (*ip_hdr).dst_addr = backend.daddr.to_be();
-    }
-    // DNAT the port
-    unsafe { (*tcp_hdr).dest = (backend.dport as u16).to_be() };
-
     if (ctx.data() + EthHdr::LEN + Ipv4Hdr::LEN) > ctx.data_end() {
         info!(&ctx, "Iphdr is out of bounds");
         return Ok(TC_ACT_OK);
     }
-
-    let mut lb_mapping = LoadBalancerMapping {
-        backend,
-        backend_key,
-        tcp_state,
-    };
 
     let tcp_hdr_ref = unsafe { tcp_hdr.as_ref().ok_or(TC_ACT_OK)? };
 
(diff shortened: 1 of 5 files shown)


dataplane/ebpf/src/ingress/tcp.rs
```diff
@@ -116,21 +126,106 @@ pub fn handle_tcp_ingress(ctx: TcContext) -> Result<i32, i64> {
...
+        }
+    }
+
+    update_tcp_conns(tcp_hdr_ref, &client_key, &mut lb_mapping)?;
+
+    // inspired by https://github.com/torvalds/linux/blob/master/samples/bpf/tcbpf1_kern.c
+    // update dst_addr in the ip_hdr
+    // recalculate the checksums
+    unsafe {
+        let backend_ip = backend.daddr.to_be();
```
astoycos, line 151: Can this let binding be outside of the unsafe block?

dataplane/ebpf/src/ingress/tcp.rs
```diff
@@ -116,21 +126,106 @@ pub fn handle_tcp_ingress(ctx: TcContext) -> Result<i32, i64> {
...
+
+    // If the packet has the RST flag set, it means the connection is being terminated, so remove it
+    // from our map.
+    if tcp_hdr_ref.rst() == 1 {
+        unsafe {
+            LB_CONNECTIONS.remove(&client_key)?;
+        }
+    }
+
+    update_tcp_conns(tcp_hdr_ref, &client_key, &mut lb_mapping)?;
```
astoycos, line 145: Should this be moved to after the cksum calculation code block as it was prior to this patch?

dataplane/ebpf/src/ingress/tcp.rs
```diff
@@ -116,21 +126,106 @@ pub fn handle_tcp_ingress(ctx: TcContext) -> Result<i32, i64> {
...
+    // recalculate the checksums
+    unsafe {
+        let backend_ip = backend.daddr.to_be();
+
+        let ret = bpf_l4_csum_replace(
+            ctx.skb.skb,
+            TCP_CSUM_OFF,
+            original_daddr as u64,
+            backend_ip as u64,
+            IS_PSEUDO | (mem::size_of_val(&backend_ip) as u64),
```
astoycos, line 158: What is the `IS_PSEUDO` const doing here?

dataplane/ebpf/src/ingress/tcp.rs
```diff
@@ -116,21 +126,106 @@ pub fn handle_tcp_ingress(ctx: TcContext) -> Result<i32, i64> {
...
+        );
+        if ret != 0 {
+            info!(
+                &ctx,
+                "Failed to update the destination IP address in the packet header");
+            return Ok(TC_ACT_OK);
+        }
+    }
+
+    // update dest in the tcp_hdr
```
astoycos, line 196:
```suggestion
    // update destination port in the tcp_hdr
```

dataplane/ebpf/src/ingress/tcp.rs
```diff
@@ -116,21 +126,106 @@ pub fn handle_tcp_ingress(ctx: TcContext) -> Result<i32, i64> {
...
+    // recalculate the checksums
+    unsafe{
+        let backend_port = (backend.dport as u16).to_be();
```
astoycos, line 199: Again lets leave as much as we can outside of the `unsafe` block

dataplane/ebpf/src/ingress/udp.rs
```diff
@@ -95,22 +104,89 @@ pub fn handle_udp_ingress(ctx: TcContext) -> Result<i32, i64> {
...
-    // TODO(astoycos) use l3_cksum_replace instead
-    unsafe { (*ip_hdr).check = 0 };
-    let full_cksum = unsafe {
-        bpf_csum_diff(
-            mem::MaybeUninit::zeroed().assume_init(),
+    // inspired by https://github.com/torvalds/linux/blob/master/samples/bpf/tcbpf1_kern.c
+    // update dst_addr in the ip_hdr
+    // recalculate the checksums
+    unsafe {
+        let backend_ip = backend.daddr.to_be();
```
astoycos, line 111: Can you condense all of this into a function in `until.rs` that can be called in both places `ingress/udp.rs` and `ingress/tcp.rs` rather than duplicating this logic?

astoycos (commented):
Looking good @tzssangglass Thanks so much for doing this! Just some questions and comments, also looks like you need to run `cargo fmt`

Otherwise my general comments are
- Try and keep the unsafe closures as minimal as possible
- Can we condense alot of this code into a separate helper function with a good comment describing what it does? Something like (`packet_redirect_cksum_update`) or something like that?

tzssangglass:
ok, I will update based on the comments.

dataplane/ebpf/src/ingress/tcp.rs
tzssangglass: fixed

dataplane/ebpf/src/ingress/tcp.rs
tzssangglass: updated

dataplane/ebpf/src/ingress/udp.rs
tzssangglass: now I have moved these functions to `utils.rs`.

dataplane/ebpf/src/ingress/tcp.rs
tzssangglass: in fact, I found that I cannot put

```rust
    if tcp_hdr_ref.rst() == 1 {
        unsafe {
            LB_CONNECTIONS.remove(&client_key)?;
        }
    }

    let mut lb_mapping = LoadBalancerMapping {
        backend,
        backend_key,
        tcp_state,
    };

    update_tcp_conns(tcp_hdr_ref, &client_key, &mut lb_mapping)?;
```

behind 

```rust
    let backend_ip = backend.daddr.to_be();
    let ret = set_ipv4_ip_dst(&ctx, TCP_CSUM_OFF, &original_daddr, backend_ip);
    if ret != 0 {
        return Ok(TC_ACT_OK);
    }

    let backend_port = (backend.dport as u16).to_be();
    let ret = set_ipv4_dest_port(&ctx, TCP_CSUM_OFF, &original_dport, backend_port);
    if ret != 0 {
        return Ok(TC_ACT_OK);
    }
```

otherwise eBPF program would load failed, eBPF verifier shows that access out of bounds

dataplane/ebpf/src/ingress/tcp.rs
tzssangglass: fixed

dataplane/ebpf/src/ingress/tcp.rs
tzssangglass: According to my understanding, `IS_PSEUDO | (mem::size_of_val(&backend_ip) as u64)` as the last argument of `bpf_l4_csum_replace()`, it can simultaneously set the IS_PSEUDO flag and specify the size of the field to be updated, allowing the function to correctly recalculate the checksum of the transport layer and consider the impact of the pseudo header.

see more:

1. result of `IS_PSEUDO | (mem::size_of_val(&backend_ip) as u64)`: `00010000 | 0100` -> `00010100`

2. in [`bpf_l4_csum_replace()`](https://elixir.bootlin.com/linux/v5.18.19/source/net/core/filter.c#L1935) 

	```c
		switch (flags & BPF_F_HDR_FIELD_MASK) {
		case 0:
			if (unlikely(from != 0))
				return -EINVAL;
	
			inet_proto_csum_replace_by_diff(ptr, skb, to, is_pseudo);
			break;
		case 2:
			inet_proto_csum_replace2(ptr, skb, from, to, is_pseudo);
			break;
		case 4:
			inet_proto_csum_replace4(ptr, skb, from, to, is_pseudo);
			break;
	```
	here `BPF_F_HDR_FIELD_MASK` is
    ```c
		enum {
		BPF_F_HDR_FIELD_MASK		= 0xfULL,
		};
	```

	`flags & BPF_F_HDR_FIELD_MASK`: `00010100 & 00001111 ` -> `00000100`, and would chosse `case 4` to update l4 checksum.

dataplane/ebpf/src/ingress/tcp.rs
tzssangglass: I think accessing `tcp_hdr` after `bpf_redirect_neigh` will cause an out-of-bounds error, but I don't know why the original code did not have this issue.

shaneutt:
I will try to review this soon
/hold

shaneutt:
@tzssangglass needs a rebase on `main`

@astoycos when you're done with your review, if you'd like me to take a pass as well please do ping me

tzssangglass:
> needs a rebase on `main`

done

astoycos (approved):
/lgtm
/approve

Huge improvement here thanks and sorry for the slowness @tzssangglass !!!

tzssangglass:
oops, it looks like k8s-ci-robot didn't squash the commit.

PR #261, opened 2024-07-16
fix dataplane and udp server builds on arm64

aryan9600:


Containerfile.dataplane
```diff
@@ -7,8 +7,13 @@ RUN apt-get install --yes \
     build-essential \
     protobuf-compiler \
     pkg-config \
-    llvm-16 \
-    musl-tools
+    musl-tools \
+    wget
+
+RUN wget -O /tmp/llvm.sh https://apt.llvm.org/llvm.sh
```
shaneutt, line 13: Switching from pulling llvm from the Debian repositories to a `curl | sudo bash` kinda situation isn't ideal, as it adds some fragility to our build system. Additionally this will always just pull the latest release, yes? I take it something in our Aya dependencies changed to require the newer LLVM, right? Is there some other image base we could use that would include a compatible version of LLVM in the package repositories?

astoycos: Is llvm-18 not packaged on ubuntu slim-bookworm?

EandrewJones: Looks like you can specify the version number with the shell script:
> For convenience there is an automatic installation script available that installs LLVM for you.
To install the latest stable version:
>
> ```sh
> bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
> ```
>
> To install a specific version of LLVM:
> 
> ```sh
> wget https://apt.llvm.org/llvm.sh
> chmod +x llvm.sh
> sudo ./llvm.sh <version number>
> ```
Looks there's an [official debian release](https://apt.llvm.org/) for 18  (under qualification branch): `apt-get install llvm-18`

Is there a reason we can't use that @aryan9600?

aryan9600: i looked around (not exhaustingly) and i coulnd't find something. i'll take another look and report back.

shaneutt: @aryan9600 heya :wave: just checking in?

shaneutt: ping

Gladdstone:
this needs to be llvm-19, now 😔

aryan9600:
this is ready for another round of review @shaneutt  @EandrewJones

EandrewJones:
@aryan9600 Still failing 😢 
image

From the logs it looks like the arch might be getting set incorrectly. Do you want it to be aarch64 or x86_64 for CI testing? The latter, correct? Well where is TARGETARCH being set?

shaneutt (commented):
If this isn't going to be super straightforward (I say this because it seems like we've had some issues) then I'm almost inclined to say we should wait until someone else comes along and requests arm64 builds before we spend more time on this?

EandrewJones:
Tell that to Flynn ;)

But I agree. arm64 builds are notoriously painful to maintain, in my
experience.

Best

Evan Jones
Website: www.ea-jones.com

shaneutt:
Oh I didn't realize Flynn needed this (the description was empty! there wasn't an issue for it!). I'll go chat with him about it, if this unblocks him then I would be in favor of pressing on actually, as I don't want anyone who's interested in the project to be "locked out" because of their system architecture if we can avoid it within a reasonable amount of effort.

EandrewJones:
My understanding is @aryan9600 initially hopped on this because Flynn
mentioned the build was broken for him. And Sanskar has been chasing llvm
versions ever since. We should confirm.

Best

Evan Jones
Website: www.ea-jones.com

shaneutt:
So yeah I talked with Flynn and all his development/testing is done on ARM64 systems so unblocking a community member is a great reason to continue. I did realize however that without the finished Rust control-plane the value of _any_ builds is a bit dubious so just keep that in mind (but otherwise godspeed).

.github/workflows/build.yaml
```diff
@@ -51,3 +51,25 @@ jobs:
     - name: Run Tests
       run: |
         make test
+  docker-build:
```
shaneutt, line 54: Since this build is currently so time consuming we should try to limit when it runs as much as feasible. At minimum, we shouldn't run this when only control-plane components are changed. At best, we should identify the _exact_ circumstances under which this _must_ be run and configure the runs as surgically as possible.

shaneutt (approved):
/approve
/lgtm

PR #276, opened 2024-09-03
new implements for getting ifindex by dst ipv4

### LINKED ISSUE #49: Dataplane interface resolution with syscalls
## Problem Statement

During #41 we added a mechanism to identify the [kind-net](https://kind.sigs.k8s.io/) network interface that would be needed for our TC program to redirect packets to `Pods`, but as a stop-gap as the PR was already getting quite large we implemented that using the `iproute2` command line. The purpose of this task is to replace that with syscalls to make it more portable.

## Prerequisites

- [x] #41 

## Additional Notes

Investigate ~https://github.com/redhat-et/emerita/blob/ade6ca9c1215704da6ce033e0b4e8a5f39dbd908/emerita/src/netlink.rs#L46~ https://github.com/rust-netlink

Check out this crate, which is higher level and doesn't require direct syscalls: https://crates.io/crates/netlink-packet-route

## Acceptance Criteria

- [ ] `iproute2` should no longer be required in order for the dataplane to work

maborosii:
fix #49

### DIFF
--- a/dataplane/api-server/src/netutils.rs
+++ b/dataplane/api-server/src/netutils.rs
@@ -5,9 +5,7 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
 */
 
 use anyhow::Error;
-use netlink_packet_core::{
-    NetlinkHeader, NetlinkMessage, NetlinkPayload, NLM_F_DUMP_FILTERED, NLM_F_REQUEST,
-};
+use netlink_packet_core::{NetlinkHeader, NetlinkMessage, NetlinkPayload, NLM_F_REQUEST};
 use netlink_packet_route::{
     route::{RouteAddress, RouteAttribute, RouteFlags, RouteHeader, RouteMessage},
     AddressFamily, RouteNetlinkMessage,
@@ -15,14 +13,18 @@ use netlink_packet_route::{
 use netlink_sys::{protocols::NETLINK_ROUTE, Socket, SocketAddr};
 use std::net::Ipv4Addr;
 
+const ERR_NO_IFINDEX: &str = "no ifindex found to route";
+const ERR_PACKET_CONSTRUCTION: &str = "construct packet failed";
+
 /// Returns an network interface index for a Ipv4 address (like the command `ip route get to $IP`)
 pub fn if_index_for_routing_ip(ip_addr: Ipv4Addr) -> Result<u32, Error> {
-    let mut socket = Socket::new(NETLINK_ROUTE)?;
-    let _port_number = socket.bind_auto()?.port_number();
+    let socket = Socket::new(NETLINK_ROUTE)?;
     socket.connect(&SocketAddr::new(0, 0))?;
 
     let mut nl_hdr = NetlinkHeader::default();
-    nl_hdr.flags = NLM_F_REQUEST | NLM_F_DUMP_FILTERED;
+
+    // NNLM_F_REQUEST: Must be set on all request messages
+    nl_hdr.flags = NLM_F_REQUEST;
 
     // construct RouteMessage
     let route_header = RouteHeader {
@@ -37,9 +39,7 @@ pub fn if_index_for_routing_ip(ip_addr: Ipv4Addr) -> Result<u32, Error> {
     route_message.attributes = vec![route_attribute];
     route_message.header = route_header;
 
-    let no_ifindex_err = format!("no ifindex found to route {}", ip_addr);
-    let con_packet_err = "construct packet failed".to_string();
-
+    // construct a message packet for netlink and serialize it to send it over the socket
     let mut packet = NetlinkMessage::new(
         nl_hdr,
         NetlinkPayload::from(RouteNetlinkMessage::GetRoute(route_message)),
@@ -48,30 +48,29 @@ pub fn if_index_for_routing_ip(ip_addr: Ipv4Addr) -> Result<u32, Error> {
     let mut buf = vec![0; packet.header.length as usize];
     // check packet
     if buf.len() != packet.buffer_len() {
-        return Err(Error::msg(con_packet_err));
+        return Err(Error::msg(ERR_PACKET_CONSTRUCTION));
     }
     packet.serialize(&mut buf[..]);
+
+    // send the serialized netlink message packet over the socket
     socket.send(&buf[..], 0)?;
 
-    let mut receive_buffer = vec![0; 4096];
-    socket.recv(&mut &mut receive_buffer[..], 0)?;
-
-    let bytes = &receive_buffer[..];
+    // read all returned messages at once
+    let (raw_netlink_message, _) = socket.recv_from_full()?;
+    let recv_route_message =
+        <NetlinkMessage<RouteNetlinkMessage>>::deserialize(&raw_netlink_message)?;
 
     // extract returned RouteNetLinkMessage
-    let (_, payload) = <NetlinkMessage<RouteNetlinkMessage>>::deserialize(bytes)?.into_parts();
-    match payload {
-        NetlinkPayload::InnerMessage(RouteNetlinkMessage::NewRoute(v)) => {
-            if let Some(RouteAttribute::Oif(idex_if)) = v
-                .attributes
-                .iter()
-                .find(|attr| matches!(attr, RouteAttribute::Oif(_)))
-            {
-                return Ok(*idex_if);
-            }
-            Err(Error::msg(no_ifindex_err.clone()))
+    if let NetlinkPayload::InnerMessage(RouteNetlinkMessage::NewRoute(message)) =
+        recv_route_message.payload
+    {
+        if let Some(RouteAttribute::Oif(idex_if)) = message
+            .attributes
+            .iter()
+            .find(|attr| matches!(attr, RouteAttribute::Oif(_)))
+        {
+            return Ok(*idex_if);
         }
-
-        _ => Err(Error::msg(no_ifindex_err)),
     }
+    Err(Error::msg(format!("{} {}", ERR_NO_IFINDEX, ip_addr)))
 }
--- a/dataplane/api-server/src/netutils.rs
+++ b/dataplane/api-server/src/netutils.rs
@@ -15,19 +15,11 @@ use netlink_packet_route::{
 use netlink_sys::{protocols::NETLINK_ROUTE, Socket, SocketAddr};
 use std::net::Ipv4Addr;
 
-// pub fn if_nametoindex(ifname: String) -> Result<u32, Error> {
-//     let ifname_c = CString::new(ifname)?;
-//     let ifindex = unsafe { libc_if_nametoindex(ifname_c.into_raw()) };
-//     Ok(ifindex)
-// }
-
-/// Returns an net interface index for a Ipv4 address, like `ip route get to`
+/// Returns an network interface index for a Ipv4 address (like the command `ip route get to $IP`)
 pub fn if_index_for_routing_ip(ip_addr: Ipv4Addr) -> Result<u32, Error> {
-    // run the linux command "ip route" to get the device's index responsible for
-    // routing the given Ipv4 address.
-    let mut socket = Socket::new(NETLINK_ROUTE).unwrap();
-    let _port_number = socket.bind_auto().unwrap().port_number();
-    socket.connect(&SocketAddr::new(0, 0)).unwrap();
+    let mut socket = Socket::new(NETLINK_ROUTE)?;
+    let _port_number = socket.bind_auto()?.port_number();
+    socket.connect(&SocketAddr::new(0, 0))?;
 
     let mut nl_hdr = NetlinkHeader::default();
     nl_hdr.flags = NLM_F_REQUEST | NLM_F_DUMP_FILTERED;
@@ -45,10 +37,8 @@ pub fn if_index_for_routing_ip(ip_addr: Ipv4Addr) -> Result<u32, Error> {
     route_message.attributes = vec![route_attribute];
     route_message.header = route_header;
 
-    let no_device_err: String = format!("no device found to route {}", ip_addr);
-    let con_packet_err: String = "construct packet failed".to_string();
-    let nl_send_msg_err: String = "netlink send message failed".to_string();
-    let nl_recv_msg_err: String = "netlink receive message failed".to_string();
+    let no_ifindex_err = format!("no ifindex found to route {}", ip_addr);
+    let con_packet_err = "construct packet failed".to_string();
 
     let mut packet = NetlinkMessage::new(
         nl_hdr,
@@ -61,21 +51,15 @@ pub fn if_index_for_routing_ip(ip_addr: Ipv4Addr) -> Result<u32, Error> {
         return Err(Error::msg(con_packet_err));
     }
     packet.serialize(&mut buf[..]);
-    socket
-        .send(&buf[..], 0)
-        .map_err(|_| Error::msg(nl_send_msg_err))?;
+    socket.send(&buf[..], 0)?;
 
     let mut receive_buffer = vec![0; 4096];
-    socket
-        .recv(&mut &mut receive_buffer[..], 0)
-        .map_err(|_| Error::msg(nl_recv_msg_err))?;
+    socket.recv(&mut &mut receive_buffer[..], 0)?;
 
     let bytes = &receive_buffer[..];
 
     // extract returned RouteNetLinkMessage
-    let (_, payload) = <NetlinkMessage<RouteNetlinkMessage>>::deserialize(bytes)
-        .map_err(|_| Error::msg(no_device_err.clone()))?
-        .into_parts();
+    let (_, payload) = <NetlinkMessage<RouteNetlinkMessage>>::deserialize(bytes)?.into_parts();
     match payload {
         NetlinkPayload::InnerMessage(RouteNetlinkMessage::NewRoute(v)) => {
             if let Some(RouteAttribute::Oif(idex_if)) = v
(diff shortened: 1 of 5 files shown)


dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,89 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
-        .ok_or_else(|| Error::msg(match_err))?
-        .as_str()
-        .to_owned();
-
-    Ok(device)
+// pub fn if_nametoindex(ifname: String) -> Result<u32, Error> {
+//     let ifname_c = CString::new(ifname)?;
+//     let ifindex = unsafe { libc_if_nametoindex(ifname_c.into_raw()) };
+//     Ok(ifindex)
+// }
```
shaneutt, line 22: It looks as though this can be removed?

```suggestion
```

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,89 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+
+/// Returns an net interface index for a Ipv4 address, like `ip route get to`
```
shaneutt, line 24:
```suggestion
/// Returns an network interface index for a Ipv4 address (like the command `ip route get to $IP`)
```

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,89 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+pub fn if_index_for_routing_ip(ip_addr: Ipv4Addr) -> Result<u32, Error> {
+    // run the linux command "ip route" to get the device's index responsible for
+    // routing the given Ipv4 address.
```
shaneutt, line 27: This leftover comment doesn't appear to be necessary now, since we're no longer using the `ip` program with `Command`:

```suggestion
```

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,89 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+    let mut socket = Socket::new(NETLINK_ROUTE).unwrap();
+    let _port_number = socket.bind_auto().unwrap().port_number();
+    socket.connect(&SocketAddr::new(0, 0)).unwrap();
```
shaneutt, line 30: We should try to avoid all these unwraps, and instead return an error result.

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,89 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+    };
+    let route_attribute = RouteAttribute::Destination(RouteAddress::Inet(ip_addr));
+    let mut route_message = RouteMessage::default();
+    route_message.attributes = vec![route_attribute];
+    route_message.header = route_header;
+
+    let no_device_err: String = format!("no device found to route {}", ip_addr);
+    let con_packet_err: String = "construct packet failed".to_string();
+    let nl_send_msg_err: String = "netlink send message failed".to_string();
+    let nl_recv_msg_err: String = "netlink receive message failed".to_string();
```
shaneutt, line 51: While it will probably be a bit more code, it may be ultimately more idiomatic to provide a custom error type for these conditions? Actually it occurs to me, does the netlink library not have custom errors that cover these conditions? :thinking:

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,89 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+                .find(|attr| matches!(attr, RouteAttribute::Oif(_)))
+            {
+                return Ok(*idex_if);
+            }
+            Err(Error::msg(no_device_err.clone()))
+        }
+
+        _ => Err(Error::msg(no_device_err)),
+    }
 }
```
shaneutt, line 93: It would ultimately be good to add tests here, albeit it's pretty obvious given the use of netlink here that tests would be relegated to something more in the integration testing kind of space. We'll need to continue thinking about how we should do testing for components like this :thinking:

dataplane/api-server/src/server.rs
```diff
@@ -99,16 +99,11 @@ impl Backends for BackendService {
         let ip = pod.ip;
         let ip_addr = std::net::Ipv4Addr::from(ip);
 
-        let device = match if_name_for_routing_ip(ip_addr) {
+        let ifindex = match if_index_for_routing_ip(ip_addr) {
             Ok(device) => device,
```
shaneutt, line 103: For consistency, perhaps?

```suggestion
            Ok(ifindex) => ifindex,
```

shaneutt (changes requested):
Looks great, thanks for sending this in! :+1: 

Mostly minor comments, and an open question about testing (which doesn't necessarily have to be solved here).

dataplane/api-server/src/netutils.rs
maborosii: so now I just add unit test for this func?

maborosii: > It would ultimately be good to add tests here, albeit it's pretty obvious given the use of netlink here that tests would be relegated to something more in the integration testing kind of space. We'll need to continue thinking about how we should do testing for components like this 🤔

so now I just add unit test for this func?

dataplane/api-server/src/netutils.rs
maborosii: When I modified this, I realized that there are only two remaining error types. It seems a bit excessive to create a custom error type just for these two. (sorry for making so many sloppy mistakes, and thank you for your patience.)

shaneutt: No worries at all. Have you confirmed whether netlink has some types already in place to cover these error conditions? If they don't and we just have the two, then it's probably fine to keep these as simple error message strings for now and we can maybe even improve this in the netlink library later.

dataplane/api-server/src/netutils.rs
shaneutt: It's difficult to make effective tests here, since we're basically just exercising netlink :thinking: Actually so that we don't have to add a long tail to this PR, and given the lack of risk to the project because the project is effectively an experimental sandbox, I think I'll back up and take a holistic look at testing for a later iteration. We can move forward as-is here and I'll follow up.

dataplane/api-server/src/netutils.rs
maborosii: yep,I don't find some error type in netlink to cover these two.

maborosii:
sorry, I don't know why integration-test failed, it passed when I rerun this patch in my downstream repo

shaneutt:
> sorry, I don't know why integration-test failed, it passed when I rerun this patch in my downstream repo

It was a flake in the Golang control-plane. We haven't tried to fix that flake because we're replacing that control-plane actively, sorry about that I re-kicked it and we're good now!

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,73 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+    packet.finalize();
+    let mut buf = vec![0; packet.header.length as usize];
+    // check packet
+    if buf.len() != packet.buffer_len() {
+        return Err(Error::msg(con_packet_err));
+    }
+    packet.serialize(&mut buf[..]);
+    socket.send(&buf[..], 0)?;
+
+    let mut receive_buffer = vec![0; 4096];
```
shaneutt, line 56: We should add a comment about why we use this size for the buffer, and ideally create a documented constant for it with a clear name.

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,73 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+    socket.recv(&mut &mut receive_buffer[..], 0)?;
```
shaneutt, line 57: Note that the `recv()` method here is a somewhat light wrapper around [libc recv](https://www.man7.org/linux/man-pages/man2/recv.2.html), so I _believe_ we'll actually want to use `MSG_TRUNC` here so that we don't accidentally drop bytes from a response. For similar reasons, while it's good that we're checking the `Result` here for `Err`, I believe we'll want to check the value returned as well :thinking:

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,73 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+
+    let bytes = &receive_buffer[..];
```
shaneutt, line 59: tiny thing, but since we only use this once from here on, perhaps we don't need the extra variable and can just hand the `receive_buffer` to `deserialize`?

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,73 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+
+    // extract returned RouteNetLinkMessage
+    let (_, payload) = <NetlinkMessage<RouteNetlinkMessage>>::deserialize(bytes)?.into_parts();
+    match payload {
+        NetlinkPayload::InnerMessage(RouteNetlinkMessage::NewRoute(v)) => {
+            if let Some(RouteAttribute::Oif(idex_if)) = v
```
shaneutt, line 65:
```suggestion
        NetlinkPayload::InnerMessage(RouteNetlinkMessage::NewRoute(message)) => {
            if let Some(RouteAttribute::Oif(idex_if)) = message
```

shaneutt (changes requested):
Second review pass!

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,73 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
-        .ok_or_else(|| Error::msg(match_err))?
-        .as_str()
-        .to_owned();
-
-    Ok(device)
+/// Returns an network interface index for a Ipv4 address (like the command `ip route get to $IP`)
+pub fn if_index_for_routing_ip(ip_addr: Ipv4Addr) -> Result<u32, Error> {
+    let mut socket = Socket::new(NETLINK_ROUTE)?;
+    let _port_number = socket.bind_auto()?.port_number();
+    socket.connect(&SocketAddr::new(0, 0))?;
```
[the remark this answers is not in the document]
shaneutt: Actually, it _could_ be a while before they respond so if you just wanna make this change and check if the tests pass, we can safely make more changes later if we learn something new with https://github.com/rust-netlink/netlink-sys/pull/23

shaneutt (commented):
So after we've resolved the above minor stuff, I think we're really close to LGTM. After that I think all that's left will be digging deeper into how we're performing the socket `send()` and `recv()` as I think some tweaking will be needed there. Hit the re-request review button to grab my attention anytime!

maborosii:
After your second review, I realized my approach was too simplistic and not rigorous enough. I need some time to dive deeper into the underlying packet structure of the `RouteMessage` response in `netlink` and properly define its data boundaries. I will submit the relevant patch after this weekend (or next week). Once again, thank you for your attention to detail.

dataplane/api-server/src/netutils.rs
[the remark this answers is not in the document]
maborosii: I initially noticed that the port_number was redundant, but since it was retained in best practices, I kept it as well. Now, I will remove it and run some local tests.

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,73 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+        table: RouteHeader::RT_TABLE_MAIN,
+        ..Default::default()
+    };
+    let route_attribute = RouteAttribute::Destination(RouteAddress::Inet(ip_addr));
+    let mut route_message = RouteMessage::default();
+    route_message.attributes = vec![route_attribute];
+    route_message.header = route_header;
+
+    let no_ifindex_err = format!("no ifindex found to route {}", ip_addr);
+    let con_packet_err = "construct packet failed".to_string();
```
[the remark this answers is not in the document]
maborosii: At first, I also wanted to initialize RouteMessage all at once, but the compiler informed me that RouteMessage is marked as non-exhaustive, and this is not allowed.

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,73 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+        address_family: AddressFamily::Inet,
+        flags: RouteFlags::LookupTable,
+        destination_prefix_length: 32,
+        table: RouteHeader::RT_TABLE_MAIN,
+        ..Default::default()
+    };
+    let route_attribute = RouteAttribute::Destination(RouteAddress::Inet(ip_addr));
+    let mut route_message = RouteMessage::default();
+    route_message.attributes = vec![route_attribute];
+    route_message.header = route_header;
```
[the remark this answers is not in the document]
maborosii: At first, I also wanted to initialize RouteMessage all at once, but the compiler informed me that RouteMessage is marked as non-exhaustive, and this is not allowed.

dataplane/api-server/src/netutils.rs
maborosii: I've only kept the size from the netlink example code. After your reminder, I captured packets and established the underlying data structure. The relevant message structure can be found at https://github.com/maborosii/netlink-capture. I am currently considering adjusting it to 128 or 256 to save space.

shaneutt:
> After your second review, I realized my approach was too simplistic and not rigorous enough. I need some time to dive deeper into the underlying packet structure of the `RouteMessage` response in `netlink` and properly define its data boundaries. I will submit the relevant patch after this weekend (or next week). Once again, thank you for your attention to detail.

Ok thanks for the heads up! Take your time, no rush, and reach out if you're in need of support. :vulcan_salute:

maborosii:
1. In the current scenario, typically only the best route is returned. In this case, should I poll the recv_buffer and traverse it until an end marker is found, or is it enough to read everything in one go?

2. Following up on question 1, should we develop a more generic solution, or only target the specific scenario we currently need?

3. If a more general solution is required, I found that https://github.com/rust-netlink/rtnetlink offers higher-level APIs. However, it doesn't directly support querying routes by target IP (see example code: https://github.com/rust-netlink/rtnetlink/blob/main/examples/get_route_kernel_filter.rs). It’s possible to achieve this by using some lower-level functions from the crate, but I'm hesitant because this approach feels somewhat messy and inconsistent.

Looking forward to your advice!

shaneutt:
Thanks for continuing on this @maborosii!

> 1. In the current scenario, typically only the best route is returned. In this case, should I poll the recv_buffer and traverse it until an end marker is found, or is it enough to read everything in one go?

Traverse the buffer to the end, I would think. When you looked over the higher level APIs, did you notice whether this is not what they do?

> 2. Following up on question 1, should we develop a more generic solution, or only target the specific scenario we currently need?

Targeted like this is _completely fine for now_. If we find a deeper need for netlink later, then we could consider refactoring into a more generic approach.

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,98 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+            }
+
+            offset += rx_packet.header.length as usize;
+            if offset == size || rx_packet.header.length == 0 {
+                offset = 0;
+                break 'buffer;
+            }
+        }
+    }
+    Err(Error::msg(format!("{} {}", ERR_NO_IFINDEX, ip_addr)))
```
shaneutt, line 101: The effort to handle the socket receive here is appreciated, but I started to note that this has become slightly complex. I started to suspect there were some higher level solutions available, and after reviewing the `Socket` documentation, I think we may be able to use `recv_from_full()`, which might enable something more like this:

```rust
    let (raw_netlink_message, _) = socket.recv_from_full()?;
    let route_message = <NetlinkMessage<RouteNetlinkMessage>>::deserialize(&raw_netlink_message)?;
```

I haven't tested this myself yet, but on paper this looks like we can make this area simpler and focus more on the error handling aspects. Let me know your thoughts? :thinking:

shaneutt (changes requested):

dataplane/api-server/src/netutils.rs
```diff
@@ -5,59 +5,98 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
-        .to_owned();
-
-    Ok(device)
+const ERR_NO_IFINDEX: &str = "no ifindex found to route";
+const ERR_PACKET_CONSTRUCTION: &str = "construct packet failed";
+
+/// Returns an network interface index for a Ipv4 address (like the command `ip route get to $IP`)
+pub fn if_index_for_routing_ip(ip_addr: Ipv4Addr) -> Result<u32, Error> {
+    let mut socket = Socket::new(NETLINK_ROUTE)?;
+    socket.bind_auto()?;
```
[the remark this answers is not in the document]
maborosii: yep, when using connect will be performed automatically, I'll remove it.

dataplane/api-server/src/netutils.rs
maborosii: The function you provided works well. I think I fell into a misconception, always trying to handle the packets received by recv and dealing with boundary issues by determining the packet type myself. I never thought that Netlink already provides more advanced methods to handle this specific scenario.

shaneutt (approved):
Awesome. In the future as we discussed there may be some room to use a more generic approach and/or more higher level rust-netlink constructs but this is a solid base to move forward with now. Thank you very much for the effort on this!

/approve
/lgtm

Issue #81, opened 2023-03-13
Add conformance tests for UDPRoute & TCPRoute


shaneutt:
Part of the purpose of this project was to help mature `UDPRoute` and `TCPRoute` in upstream [Gateway API](https://gateway-api.sigs.k8s.io/).

Part of that maturity is a need for conformance tests which multiple implementations can subscribe to. The purpose of this task is to add the conformance test suite for Gateway API to this repository, resolve https://github.com/kubernetes-sigs/gateway-api/issues/1792 and then include those tests here.

This should inherently resolve the following:

- [ ] https://github.com/kubernetes-sigs/gateway-api/issues/1794
- [ ] https://github.com/kubernetes-sigs/gateway-api/issues/1792

shaneutt:
I'll be taking this on in support of https://github.com/kubernetes-sigs/gateway-api/issues/2035

shaneutt:
Note that this is now blocked on https://github.com/kubernetes-sigs/blixt/milestone/8.

cnvergence:
Recently I have added the initial UDPRoute test in https://github.com/kubernetes-sigs/gateway-api/pull/2661.
I would like to continue the work on adding more scenarios and verifying the support on blixt as well.

cnvergence:
hi @shaneutt, sorry, I needed to drop off for the last few months.
I saw great progress in moving the controller to Rust.
Let me take a look this week by moving UDPRoute test here.

shaneutt:
No worries whatsoever! Thanks for pinging, if you need any support hit us up in `#blixt` on Kubernetes Slack! :vulcan_salute:

cnvergence:
I have ran into some issues, when running tests
```bash
blixt-controlplane-664b9f96f6-6nnxd manager 2024-08-29T10:01:45Z	INFO	Observed a panic in reconciler: runtime error: invalid memory address or nil pointer dereference	{"controller": "udproute", "controllerGroup": "gateway.networking.k8s.io", "controllerKind": "UDPRoute", "UDPRoute": {"name":"udp-coredns","namespace":"gateway-conformance-infra"}, "namespace": "gateway-conformance-infra", "name": "udp-coredns", "reconcileID": "0c314b36-5a47-41ff-b898-0f2483a3b829"}
blixt-controlplane-664b9f96f6-6nnxd manager panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[...]

PR #277, opened 2024-09-03
Bump gateway-api to v1.1.0

cnvergence:
As mentioned in https://github.com/kubernetes-sigs/blixt/issues/81, we should bump the gw-api version to the latest


shaneutt:
Hey @cnvergence! We appreciate you jumping on this. For some reason when we talked about this in #81 I had it in my head that we were talking about updating the Gateway API version in the Rust code, not the Golang control plane. We [are in the midst of replacing the Golang code](https://github.com/kubernetes-sigs/blixt/milestone/8), so actually if this is not being as straightforward as it should be, maybe we could just drop it for now? I'm sorry for the crossed wires, this has actually got me thinking that maybe now is the time to just remove the old Golang control-plane to reduce confusion going forward :thinking:

shaneutt:
Spoke with @cnvergence in Slack (thanks for reaching out!) we're going to close this for now given the above context. Next steps to follow up are to finish the control-plane rewrite, and then we should be able to get back to updates and conformance tests for Gateway API :rocket:

PR #280, opened 2024-09-08
feat(mTLS): adds mTLS support to dataplane api-server

### LINKED ISSUE #50: mTLS for dataplane API
## Problem Statement

We want to have [mutual TLS authentication](https://grpc.io/docs/guides/auth/) for the dataplane GRPC API.

## Acceptance Criteria

-  [ ] mTLS is automatically configured for encrypting _and authenticating_ controlplane to dataplane communications

EandrewJones:
Closes #50.

### Questions
I am not sure how best to add tests for this. Unit testing the config trivially tests the behavior of Opt which we know works. Unit testing `setup_tls` forces me to mock `read_to_string` in the absence of an actual file to read in and all the ways I've read about doing that in rust feel really kludgie / dirty up the code with dependency injections, etc.

1. Is this all we need to add mTLS support for the api-server?
2. Is there some sort of API contract we need to maintain for this such that I can write a conformance or integration test?
3. What documentation would we like to see for this?


dataplane/api-server/src/config.rs
```diff
@@ -0,0 +1,61 @@
...
+use clap::Parser;
+use std::path::PathBuf;
+
+#[derive(Debug, Parser, Clone)]
+pub struct GrpcConfig {
+    #[clap(long, default_value = "false")]
+    pub enable_tls: bool,
+    #[clap(long, default_value = "false")]
+    pub enable_mtls: bool,
+    pub certificate_authority_root_path: Option<PathBuf>,
```
aryan9600, line 15: i might be lacking the required context but is there a reason this is not `server_certificate_authority_root_path`?

dataplane/api-server/src/config.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub server_certificate_path: Option<PathBuf>,
+    pub server_private_key_path: Option<PathBuf>,
+    pub client_certificate_authority_root_path: Option<PathBuf>,
+    pub client_certificate_path: Option<PathBuf>,
+    pub client_private_key_path: Option<PathBuf>,
+}
+
+impl GrpcConfig {
+    pub fn validate(&self) -> Result<(), String> {
+        fn validate_paths(paths: &[(&Option<PathBuf>, &str)]) -> Result<(), String> {
```
aryan9600, line 25: is this solely done to avoid a move? asking because i'm not a big fan of inline functions unless they're essential. this can be just its own function that accepts a vector. please correct me if i'm wrong.

dataplane/api-server/src/lib.rs
```diff
@@ -5,33 +5,78 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+    if tls_config.enable_tls || tls_config.enable_mtls {
+        let mut tls = ServerTlsConfig::new();
+
+        if let Some(cert_path) = &tls_config.server_certificate_path {
+            let cert = fs::read_to_string(cert_path).expect("Error reading server certificate");
+            let key = fs::read_to_string(
+                tls_config
+                    .server_private_key_path
+                    .as_ref()
+                    .expect("Missing private key path"),
```
aryan9600, line 57: lets bubble this error up instead? `async fn start()` already returns a `Result`, so we can just return it up there.

dataplane/api-server/src/config.rs
```diff
@@ -0,0 +1,61 @@
...
+    #[clap(long, default_value = "false")]
+    pub enable_tls: bool,
+    #[clap(long, default_value = "false")]
+    pub enable_mtls: bool,
+    pub certificate_authority_root_path: Option<PathBuf>,
+    pub server_certificate_path: Option<PathBuf>,
+    pub server_private_key_path: Option<PathBuf>,
+    pub client_certificate_authority_root_path: Option<PathBuf>,
+    pub client_certificate_path: Option<PathBuf>,
+    pub client_private_key_path: Option<PathBuf>,
```
aryan9600, line 20: the below two properties don't seem to be actually used anywhere?

dataplane/api-server/src/config.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+impl GrpcConfig {
+    pub fn validate(&self) -> Result<(), String> {
```
aryan9600, line 24: i couldn't find where this is being called. please ignore if clap calls this automatically for us.

dataplane/loader/src/main.rs
```diff
@@ -20,6 +21,8 @@ use log::{info, warn};
 struct Opt {
     #[clap(short, long, default_value = "lo")]
     iface: String,
+    #[clap(flatten)]
+    tls_config: GrpcConfig,
```
aryan9600, line 25: maybe?

```suggestion
    grpc_tls_config: GrpcConfig,
```

dataplane/api-server/src/config.rs
```diff
@@ -0,0 +1,61 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+*/
+use clap::Parser;
+use std::path::PathBuf;
+
+#[derive(Debug, Parser, Clone)]
+pub struct GrpcConfig {
```
aryan9600, line 10: maybe?

```suggestion
pub struct GrpcTLSConfig {
```

aryan9600:
> Unit testing setup_tls forces me to mock read_to_string in the absence of an actual file to read in

how about generating certs during the test, writing them to temp files and then passing those to `setup_tls`?

> 1. Is this all we need to add mTLS support for the api-server?

this looks good. we can test it using https://github.com/islishude/grpc-mtls-example/blob/main/cmd/client/main.go.

> 2.  Is there some sort of API contract we need to maintain for this such that I can write a conformance or integration test?

we don't have an integration test for the current grpc server. i guess an integration test for the api server would involve spinning up the server and calling each method from an independent client? for testing mTLS specifically, i think we can just get away with only testing one method, since the thing we want to test is actually the TLS connection. something, like the answer to your first question, but coded in a script.

>  3. What documentation would we like to see for this?

we don't really have any docs apart from code comments. i think code comments for now is fine. documentation is an entirely separate conversation that we need to have soon.

dataplane/api-server/src/config.rs
EandrewJones: Updating name.

dataplane/api-server/src/config.rs
EandrewJones: No moving at play here. I generally agree nested functions suck and avoid them. In this case, I felt it wasn't that big of a deal because it's not intended to be used anywhere outside of the validate method. I tend to feel extremely tight coupling is the only excuse for nested functions.

Either way, I moved it out.

dataplane/api-server/src/config.rs
EandrewJones: I swore in my reading, I saw clap would use this method for us. But, alas, I cannot find it in docs. I added an explicit call for this now.

If you know of a better way to get clap to perform validation at parse time, please let me known.

dataplane/loader/src/main.rs
EandrewJones: Good call out.

I have changed the name of the struct to match the function argument: `GrpcConfig` --> `TlsConfig`. There's nothing about the config that makes it unique to Grpc so I don't know why I named it that way originally.

dataplane/api-server/src/config.rs
EandrewJones: Changed to TLSConfig

dataplane/api-server/src/config.rs
EandrewJones: correct. These are only relevant client side or if this server also acted as a client w.r.t to a different server.

Removing.

dataplane/api-server/src/lib.rs
EandrewJones: K.

EandrewJones:
> how about generating certs during the test, writing them to temp files and then passing those to `setup_tls`?

Added.

> > 1. Is this all we need to add mTLS support for the api-server?
> 
> this looks good. we can test it using https://github.com/islishude/grpc-mtls-example/blob/main/cmd/client/main.go.

If we're not treating this as an integration test, where should we add this so it runs in our pipeline? We need the dataplane server running so we can mimic the client, so it seems like it should go into the integration test pipeline (`tests/integration`). However, the makefile states these tests are deprecated, so should I really be adding new tests to the suite?

> > 3. What documentation would we like to see for this?
> 
> we don't really have any docs apart from code comments. i think code comments for now is fine. documentation is an entirely separate conversation that we need to have soon.

Okay.

dataplane/api-server/src/config.rs
```diff
@@ -0,0 +1,51 @@
...
+
+#[derive(Debug, Parser, Clone)]
+pub struct TLSConfig {
+    #[clap(long, default_value = "false")]
+    pub enable_tls: bool,
+    #[clap(long, default_value = "false")]
+    pub enable_mtls: bool,
+    pub server_certificate_path: Option<PathBuf>,
+    pub server_private_key_path: Option<PathBuf>,
+    pub client_certificate_authority_root_path: Option<PathBuf>,
```
shaneutt, line 18: So in general I've seen a lot of `Option<T>` fields in struct attributes over the years and that's fine in general. What's interesting in this case is that you're not trying to express that these fields are ultimately optional, but instead that they are optional situationally based on the values of other fields. In effect, this means that you have a couple of different _variants_ of a `TLSConfig` that are acceptable.

Given that notion, and the fact that I immediate thought "variants", I think you should consider whether an [Enumeration](https://doc.rust-lang.org/rust-by-example/custom_types/enum.html) might actually be a better fit here, where we would have a `TLSConfig::ServerOnlyTLS` variant and a `TLSConfig::MutualTLS`.

Now... the only problem with this is that I think the standards really don't accept `ServerOnlyTLS` as official nomenclature everywhere they often just say `TLS`, so... that might require some finesse but either way just think this over and let me know what you think.

dataplane/api-server/src/config.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+fn validate_paths(paths: &[(&Option<PathBuf>, &str)]) -> Result<()> {
+    for (path, name) in paths {
+        if path.is_none() {
+            return Err(anyhow!("Missing required path: {}", name));
+        }
+    }
+    Ok(())
+}
```
shaneutt, line 51: If we go with the `Enum` approach outlined above, then none of this Validation would be needed, but we could create [associated functions](https://doc.rust-lang.org/stable/book/ch05-03-method-syntax.html#associated-functions) to [construct](https://rust-unofficial.github.io/patterns/idioms/ctor.html) these variants.

dataplane/api-server/src/config.rs
EandrewJones: Why not just `TLSConfig::TLS` and `TLSConfig::MutualTLS`? Those correspond to the `TLS` and `mTLS` acronyms, respectively, which are standard afaik.

dataplane/api-server/src/config.rs
EandrewJones: Is this even necessary? Shouldn't clap handle this for us during parsing? The only reason I was doing this was to enforce the conditional dependencies between fields. The enum renders that irrelevant.

Can't we just pass `opt.tls_config` down to the `setup_tls` function and match on the variant?

dataplane/api-server/src/config.rs
shaneutt: I'm good with those names :+1:

dataplane/api-server/src/config.rs
shaneutt: Not necessary no, just a thought depending on how you'd like this to end up looking. But the spirit behind it was in-line with the second thing you were suggesting: once we use an enum "validation" does effectively just get outmoded by matching on the variants and applying the appropriate functionality thereafter. :+1:

EandrewJones:
DO NOT MERGE. Still needs a test.

EandrewJones:
I pushed up a go test that creates certs, spins up the dataplane, and dials the grpc w/ mTLS from a client. The test passes, but I suspect it's a false positive because the dataplane exits almost immediately on my machine due to an error:

```
Error: map error: failed to create map `AYA_LOGS` with code -1

Caused by:
    0: failed to create map `AYA_LOGS` with code -1
    1: Operation not permitted (os error 1)
Error: map error: failed to create map `GATEWAY_INDEXES` with code -1

Caused by:
    0: failed to create map `GATEWAY_INDEXES` with code -1
    1: Operation not permitted (os error 1)
```
I only care about having a valid test at this point and adding it to our CI.

test/integration/dataplane_mtls_test.go
```diff
@@ -0,0 +1,182 @@
...
+		"-p", "9874:9874",
+		"-v", os.Getenv("PWD")+"/certs:/app/certs",
+		imageName,
+		"mutual-tls",
+		"--server-certificate-path", "/app/certs/server.pem",
+		"--server-private-key-path", "/app/certs/server-key.pem",
+		"--client-certificate-authority-root-path", "/app/certs/root.pem",
+	)
+	return cmd.Run()
+}
```
EandrewJones, line 162: Not using KTF to spin up the dataplane. I believe that's a problem.

shaneutt: Right, we use KTF specifically for MetalLB which we use for IPAM for our (bpf hijacked) services right now. So for a complete testing environment we're still a bit bound to KTF (or you can manually reproduce the same environment with your own metallb configmap).

EandrewJones:
Does the golang controlplane archival have any impact on this? Or do I just
need to stand up KTF before I run these tests.

Best

Evan Jones
Website: www.ea-jones.com

shaneutt:
> Does the golang controlplane archival have any impact on this? Or do I just need to stand up KTF before I run these tests?

The archival shouldn't have an impact: there's a `make build.cluster` target that sends the right flags (and will automatically install `ktf` for you). The Golang integration tests actually were left in `main` despite removing all the other Go code (since the tests in still have value and help us validate our Rust changes until we replace them with Rust tests).

EandrewJones:
@shaneutt @aryan9600 This should be ready for final review. I finally have passing tests. Upon `lgtm`, I'll squash my commits.

EandrewJones:
Rebased.

Best

Evan Jones
Website: www.ea-jones.com

EandrewJones:
Looks like changes to the docker-build on main break this branches docker-build now. Looks like it's due to the branch name breaking docker syntax (can't include '#' is my guess).

I won't have time to touch this this week. :(

shaneutt:
Ok! No worries, no rush.

EandrewJones:
You still can review. I don't recall what I've done since you last
reviewed. Maybe the enum variants and the test?

dataplane/api-server/src/lib.rs
```diff
@@ -5,33 +5,128 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+    tls_config: Option<TLSConfig>,
+) -> Result<()> {
+    // Tonic itself doesn't provide a built-in mechanism for selectively
+    // applying TLS based on routes, as TLS configuration is tied to the
+    // entire server and managed at the transport layer, not at the
+    // application layer where routes are defined.
+    //
+    // Solution: separate gRPC services
+    //
+    // Public server without mTLS for healthchecks
```
shaneutt, line 41: If it's OK with you I would like to put a finer point on this comment to make it clear that no dataplane functionality is enabled through this unencrypted endpoint:

```suggestion
    // Public server without TLS (healthchecks ONLY)
```

dataplane/api-server/src/lib.rs
```diff
@@ -5,33 +5,128 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+    });
+
+    tokio::try_join!(healthchecks, backends)?;
+
     Ok(())
 }
+
+pub fn setup_tls(mut builder: Server, tls_config: &Option<TLSConfig>) -> Result<Server> {
+    // TLS implementation drawn from Tonic examples.
+    // See: https://github.com/hyperium/tonic/blob/master/examples/src/tls_client_auth/server.rs
```
shaneutt, line 71: I always appreciate when we link to source materials and relevant external documentation, ty :+1:

dataplane/api-server/tests/test_setup_tls.rs
```diff
@@ -0,0 +1,87 @@
...
+    let builder = Server::builder();
+
+    // Run the setup_tls function and ensure no error is thrown
+    let result = setup_tls(builder, &tls_config);
+    assert!(
+        result.is_ok(),
+        "setup_tls should succeed with valid self-signed certs"
+    );
+    Ok(())
+}
```
shaneutt, line 87: For completeness it would be ideal to cover the negative inverse cases of both these positive test cases. That is to say: these tests currently build up to an expected success result and verify it, we should _also_ cover the failure cases and verify that we get the expected error result.

dataplane/loader/src/main.rs
```diff
@@ -16,12 +17,51 @@ use clap::Parser;
...
+/// ```bash
+/// # Running with default interface and no TLS config:
+/// $ dataplane
+///
+/// # Running with a specified interface and server-only TLS config:
+/// $ dataplane --iface eth0 tls --server-certificate-path /path/to/cert --server-private-key-path /path/to/key
+///
+/// # Running with mutual TLS config:
+/// $ dataplane --iface eth0 mutual-tls --server-certificate-path /path/to/cert --server-private-key-path /path/to/key --client-certificate-authority-root-path /path/to/ca
+/// ```
```
shaneutt, line 64: Your taking the time to add documentation here is greatly appreciated :bow: :+1:

dataplane/loader/src/main.rs
```diff
@@ -62,6 +102,7 @@ async fn main() -> Result<(), anyhow::Error> {
         .context("failed to attach the egress TC program")?;
 
     info!("starting api server");
+    info!("Using tls config: {:?}", &opt.tls_config);
```
shaneutt, line 105: :+1:

test/integration/dataplane_mtls_test.go
```diff
@@ -0,0 +1,118 @@
+//go:build dataplane_tests
```
shaneutt, line 1: I'm not going to say no to adding Golang tests here because this PR review has gone on long enough already and I appreciate your efforts and would prefer to get this in and focus on iterating than to try and stop and do a major rework again now. That said, I felt it would be remiss of me not to point out that we'll likely need to rewrite this into Rust tests at some point so just keep that in mind! Feel free to consider this comment resolved at your discretion.

test/integration/dataplane_mtls_test.go
```diff
@@ -0,0 +1,118 @@
...
+var (
+	CAFile         = certFile("ca.pem")
+	ClientCertFile = certFile("client.pem")
+	ClientKeyFile  = certFile("client-key.pem")
+	ServerCertFile = certFile("server.pem")
+	ServerKeyFile  = certFile("server-key.pem")
+)
+
+// TestGRPCClient tests the gRPC client against a running Docker container.
+func TestGRPCClient(t *testing.T) {
```
shaneutt, line 44: Is it worth pointing out here (for the future code reader) that this test is verifying our mTLS configuration works?

```suggestion
// TestGRPCClientWithMTLS tests the client connectivity to the dataplane's gRPC API server. Mutual TLS (mTLS)
// is required to connect to the dataplane API server, and so as a part of this test we create a client certificate
// using the same CA the server used in order to establish mutual trust.
func TestGRPCClientWithMTLS(t *testing.T) {
```

Maybe a bit too wordy, so feel free to ignore this specific suggestion, but I think a breadcrumb here that let's future readers better understand that mTLS configuration is effectively being tested here would be good.

test/integration/dataplane_mtls_test.go
```diff
@@ -0,0 +1,118 @@
...
+func certFile(filename string) string {
+	wd, _ := os.Getwd()
+	projectDir := filepath.Clean(filepath.Join(wd, "..", ".."))
+	if certsDir := os.Getenv("TEST_CERTS_PATH"); certsDir != "" {
+		return filepath.Join(projectDir, certsDir, filename)
+	}
+	panic("Env var TEST_CERTS_PATH not found. Please specify path to mTLS test certs")
+}
+
+func setupTLSConfig(cfg TLSConfig) (*tls.Config, error) {
```
shaneutt, line 76: There's a bunch of optionality present in here, in that it will try and generate certificates if not present, e.t.c.. However (if I'm not misunderstanding) we actually currently expect for all runs of this test that all values are present? If so, and because this is a private function only used here for this purpose, we may actually just want to _fail_ when things are not present instead as this will help us more quickly realize when we've broken our build system. Wdyt?

shaneutt (approved):
Thank you so much for all your hard work making this happen.

I approve in that I think we're just about ready to go and I don't see any major blocking issues, however I do have some comments I would like to see resolved before we merge please. Also we gotta figure out the CI issue, but in any case, I think we're really close. Thank you!

Makefile
```diff
@@ -216,6 +230,36 @@ lint: ## Lint Rust code
...
+test.rmcert:
+	rm $(TEST_CERTS_PATH)/{*.pem,*.csr}
+
+.PHONY: test.dataplane.integration
+test.dataplane.integration:
+	go clean -testcache
+	TEST_CERTS_PATH=./$(TEST_CERTS_PATH) \
+	BLIXT_DATAPLANE_IMAGE=$(BLIXT_DATAPLANE_IMAGE):$(TAG) \
+	GOFLAGS="-tags=dataplane_tests" go test -race -v ./test/integration/...
+	$(MAKE) test.rmcert
```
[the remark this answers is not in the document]
EandrewJones: I am happy mpt to remove them since this test will mostly run in ephemeral CI environments. But I think we _should keep_ the targets for creating them. We want to have an automated way to create self-signed certs.

dataplane/loader/src/main.rs
EandrewJones: I don't love the way rust destructures (or doesn't, rather) the config when printing it, but that's a small nit I can live with for now.

test/integration/dataplane_mtls_test.go
EandrewJones: Yeah, I'm aware. I wrote these in go out of laziness; I drew on the precedent of integration tests written in golang.

test/integration/dataplane_mtls_test.go
EandrewJones: This function does not generate certs if they are not absent, it just reads in the certs and converts them into the tls.config struct that grpc expects.

However, it doesn't abort early if paths to CertFile, KeyFile or CAFile are not present. I can add those.

EandrewJones:
@shaneutt I addressed all your comments and fixed the docker-build naming issue by switching the workflow to use the PR number instead of branch name since it's guaranteed to comply with docker's naming rules.

The last thing I need to do is squash. Is there a way for me to squash and re-push without triggering the docker-build action? It's extremely expensive/long-running.

Makefile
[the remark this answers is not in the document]
shaneutt: Right, exactly what I was getting at. Keep the _targets_ but don't do cleanup automatically here. :+1:

shaneutt (approved):
/approve
/unhold

/cc @aryan9600

shaneutt:
Oh wow yeah 70m build time... holy cow.

shaneutt (approved):
We could always spend more time here, but given the project is still nascent and we do not have end-users for breakages to impact, let's merge this and start using it!

Thanks for all your hard work @EandrewJones :bow: 

/lgtm

PR #281, opened 2024-09-13
Fix: detect when hardware platform is aarch64

Gladdstone:
Currently the makefile fails to select the correct build platform for arm devices that return `aarch64` as the result for `uname -m`.


shaneutt:
Nice, thank you!

/approve
/lgtm

PR #282, opened 2024-09-13
fix warnings when build images

nayihz:
fix warnings when build images
```
 1 warning found (use docker --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 2)
```
and
```
 => WARN: LegacyKeyValueFormat: "LABEL key=value" should be used instead of legacy "LABEL key value" format (line 5)                                                                           0.0s
 => WARN: LegacyKeyValueFormat: "LABEL key=value" should be used instead of legacy "LABEL key value" format (line 6)                                                                           0.0s
 => WARN: LegacyKeyValueFormat: "LABEL key=value" should be used instead of legacy "LABEL key value" format (line 8)  
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,10 @@
 # ------------------------------------------------------------------------------
 
 # IMAGES used when running tests.
-BLIXT_CONTROLPLANE_IMAGE ?= ghcr.io/kubernetes-sigs/blixt-controlplane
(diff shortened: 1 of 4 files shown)


nayihz:
/assign @shaneutt 
PTAL at your convenience.

PR #312, opened 2024-12-02
fix dataplane loader 

jokestax:
## Description

The error was `error cannot find value 'bpf-program' in this scope`,because the variable `bpf_program` is only available when `debug_assertions` is enabled, so fixed it by using correct variable whether pogram is in release build or debug and also the `include_bytes_aligned!` macro was unable to load the file at runtime. Replaced with `fs::read()` to load the file 

## Testing

To test this change:

1. Clone the repository.
2. Run `make build.image.dataplane`.
3. Ensure the build completes successfully without any errors.

### DIFF
--- a/dataplane/loader/src/main.rs
+++ b/dataplane/loader/src/main.rs
@@ -5,6 +5,7 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
 */
 
 use std::fs;
+use std::path::Path;
 use std::net::Ipv4Addr;
 
 use anyhow::Context;
@@ -31,17 +32,15 @@ async fn main() -> Result<(), anyhow::Error> {
 
     info!("loading ebpf programs");
 
-    let path = "../../target/bpfel-unknown-none/debug/loader";
-    let bytes = fs::read(path);
-
-    #[cfg(debug_assertions)]
-    let mut bpf_program = Ebpf::load(&bytes)?;
-    
-    #[cfg(not(debug_assertions))]
-    let mut bpf = Ebpf::load(&bytes)?;
-    if let Err(e) = EbpfLogger::init(&mut bpf_program) {
-        warn!("failed to initialize eBPF logger: {}", e);
-    }
+    let mut bpf_program = if cfg!(debug_assertion) {
+        let path = "../../target/bpfel-unknown-none/debug/loader";
+        let bytes = fs::read(Path::new(path));
+        Ebpf::load(&bytes);
+    } else {
+        let path = "../../target/bpfel-unknown-none/release/loader";
+        let bytes = fs::read(Path::new(path));
+        Ebpf::load(&byte);
+    };
 
     info!("attaching tc_ingress program to {}", &opt.iface);
 
--- a/dataplane/loader/src/main.rs
+++ b/dataplane/loader/src/main.rs
@@ -32,15 +32,14 @@ async fn main() -> Result<(), anyhow::Error> {
 
     info!("loading ebpf programs");
 
-    let mut bpf_program = if cfg!(debug_assertions) {
-        let path = "../../target/bpfel-unknown-none/debug/loader";
-        let bytes = fs::read(Path::new(path)).expect("Failed to read debug loader");
-        Ebpf::load(&bytes).expect("Failed to load eBPF program")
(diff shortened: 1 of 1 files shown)


dataplane/loader/src/main.rs
```diff
@@ -4,13 +4,15 @@ Copyright 2023 The Kubernetes Authors.
...
 
+use std::fs;
+use std::path::Path;
 use std::net::Ipv4Addr;
 
 use anyhow::Context;
 use api_server::start as start_api_server;
 use aya::maps::HashMap;
 use aya::programs::{tc, SchedClassifier, TcAttachType};
-use aya::{include_bytes_aligned, Ebpf};
```
shaneutt, line 13: Could you please tell me a bit more about the removal of the use of this macro? The implications of removing the use of this wrapper (which is just a wrapper around [the standard library include_bytes macro](https://doc.rust-lang.org/std/macro.include_bytes.html)) is that we'll no longer be enforcing the 32bit alignment of the bpf objects we load, so in _theory_ failures can occur. I'm inclined to say we should leave that, but I'm curious to hear if you have additional thoughts here? :thinking:

dataplane/loader/src/main.rs
```diff
@@ -30,14 +32,15 @@ async fn main() -> Result<(), anyhow::Error> {
...
-    let mut bpf_program = Ebpf::load(include_bytes_aligned!(
+    let path = if cfg!(debug_assertions) {
         "../../target/bpfel-unknown-none/debug/loader"
-    ))?;
-    #[cfg(not(debug_assertions))]
-    let mut bpf_program = Ebpf::load(include_bytes_aligned!(
+    } else {
         "../../target/bpfel-unknown-none/release/loader"
-    ))?;
+    };
```
shaneutt, line 39: > The error was error cannot find value 'bpf-program' in this scope,because the variable bpf_program is only available when debug_assertions is enabled

Could you please tell me a bit more about how you originally triggered this problem?

> include_bytes_aligned! macro was unable to load the file at runtime. Replaced with fs::read() to load the file

So you're quite right that on a fresh clone of the repository the object file will be missing until the xtask is run to build the BPF object, and that's probably something we can improve in our tooling. But again, I'm curious about the original steps that triggered this. Was the `Makefile` used or was it more like `cd dataplane; cargo build` that failed?

shaneutt, line 39: > The error was error cannot find value 'bpf-program' in this scope,because the variable bpf_program is only available when debug_assertions is enabled

Could you please tell me a bit more about how you originally triggered this problem?

> include_bytes_aligned! macro was unable to load the file at runtime. Replaced with fs::read() to load the file

So you're quite right that on a fresh clone of the repository the object file will be missing until the xtask is run to build the BPF object, and that's probably something we can improve in our tooling (and documentation). Again however, I'm curious about the original steps that triggered this: was the `Makefile` used or was it more like `cd dataplane; cargo build` that failed?

shaneutt (commented):
Thank you so much for your contribution @jokestax we appreciate you joining us here :wave: 

I have a couple questions about the patch, let me know your thoughts?

dataplane/loader/src/main.rs
jokestax: I encountered a scope issue when running `make build.image.dataplane`, because the variable bpf_program is only available when debug_assertions is enabled resulting in an error. However, I noticed that the PR merged today (https://github.com/kubernetes-sigs/blixt/commit/427839a319483f958001ba47fa6b073a08ee7755) has fixed the issue.

shaneutt:
@jokestax If I'm understanding correctly, the problem is resolved. Would it be reasonable to consider this closed or was there more to do here? Sorry for the bad timing with https://github.com/kubernetes-sigs/blixt/commit/427839a319483f958001ba47fa6b073a08ee7755!

jokestax:
@shaneutt ,yes, the issue is resolved. we can go ahead and close it. Thank you again for the quick response and fix :raised_hands:

shaneutt:
Oh, yep! I see you joined the Slack already cool. Yes feel free to take a stab at that one, it's a bit old: if it seems kinda crunchy or if you need support please reach out in Slack. :clinking_glasses:

jokestax:
Thanks a lot! I'll get started and reach out in Slack!

PR #341, opened 2025-01-17
chore: try to make contorlplane dev better

aryan9600:
This PR attempts to add some distinctions between components that build on Linux and components that don't, mainly in service of MacOS development environments, as historically the entire development ecosystem and build system for Blixt has been Linux-only.

Also provided are some general fixes and improvements for building and testing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,13 +1,14 @@
 [workspace]
 resolver = "2"
 members = [
+	"controlplane",
 	"dataplane/api-server",
 	"dataplane/common",
 	"dataplane/loader",
 	"tools/udp-test-server",
 	"xtask",
 ]
-exclude = ["dataplane/ebpf", "controlplane"]
+exclude = ["dataplane/ebpf"]
 
 [workspace.package]
 edition = "2021"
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -1,13 +1,14 @@
 [workspace]
 resolver = "2"
 members = [
+	"controlplane",
```
shaneutt, line 4: :+1:

dataplane/common/Cargo.toml
```diff
@@ -14,4 +14,4 @@ default = []
 user = [ "aya" ]
 
 [dependencies]
-aya = { workspace = true, optional=true }
+aya = { version = "0.13.1", optional=true }
```
shaneutt, line 17: It may be reasonable to remove the aya dependencies at the workspace level as you did, as we wouldn't expect eBPF code anywhere but the dataplane crate. That said, just note that you did leave the base `aya` dependency in the workspace Cargo.toml.

dataplane/loader/Cargo.toml
```diff
@@ -10,11 +10,11 @@ name = "loader"
 path = "src/main.rs"
 
 [dependencies]
+aya = { version = "0.13.1" , features=["async_tokio"] }
+aya-log = { version = "0.2.1" }
```
shaneutt, line 14: Now I'm kinda second guessing: was there some specific problem with having aya depds defined at the workspace level? Because now we have version configured in multiple places and I'm just curious what the impetus was

xtask/src/build_proto.rs
```diff
@@ -7,18 +7,33 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+            .compile(&[proto_file], &["."])?;
+    }
+
+    if opts.controlplane {
+        tonic_build::configure()
+            .protoc_arg("--experimental_allow_proto3_optional")
+            .build_client(true)
+            .out_dir("./controlplane/src/client")
+            .compile(&[proto_file], &["."])?;
+    }
```
shaneutt, line 36: haha, woops

xtask/src/main.rs
```diff
@@ -23,24 +25,38 @@ pub struct Options {
 
 #[derive(Debug, Parser)]
 enum Command {
+    #[cfg(target_os = "linux")]
     BuildEbpf(build_ebpf::Options),
-    BuildProto(build_proto::Options),
-    Run(run::Options),
+    #[cfg(target_os = "linux")]
     GrpcClient(grpc::Options),
```
shaneutt, line 31: So wait, wouldn't we want the grpc client build available on other OS?

dataplane/common/Cargo.toml
aryan9600: good catch!

aryan9600: i guess aya is not a linux only crate, but related crates like aya-log are.

dataplane/loader/Cargo.toml
aryan9600: its the introduction of the controlplane to the workspace. the problem with keeping aya (and netlink) as a workspace dependency is that they are compiled regardless of whether the package you're compiling uses it. having these deps in the workspace would mean that you can't build the contorlplane on macOS from the workspace root, since these creates are meant only for linux

xtask/src/main.rs
aryan9600: this disables anything on non-linux that imported any dataplane crates. upon further thinking, now that we have the bindings generated for the controlplane crate as well, we can use that crate to import the bindings in the grpc client tool.

xtask/src/build_proto.rs
```diff
@@ -7,18 +7,33 @@ SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
...
+            .build_server(true)
+            .out_dir("./dataplane/api-server/src")
+            .compile(&[proto_file], &["."])?;
+    }
+
+    if opts.controlplane {
+        tonic_build::configure()
+            .protoc_arg("--experimental_allow_proto3_optional")
+            .build_client(true)
+            .out_dir("./controlplane/src/client")
```
shaneutt, line 34: So there's a slight nuance here that I was thinking about: this is the client for the dataplane, but we're placing it under the controlplane? Would it potentially be better to place this dataplane client under `dataplane/`, even if under it's own crate?

shaneutt: Talked it over in Slack: splitting this out into its own PR. :+1:

shaneutt (approved):
/approve
/lgtm

PR #356, opened 2025-01-28
(feat): debug logs on data plane for health service addresses

buraksekili:
This PR
- adds debug and info logs to print out addresses & ports being used by data plane for gRPC health service.
- updates xtask command to run data plane with debug log level for `api_server`
- fixes incorrect xtask command name for controlplane (typo).

### DIFF
--- a/dataplane/api-server/src/lib.rs
+++ b/dataplane/api-server/src/lib.rs
@@ -45,6 +45,8 @@ pub async fn start(
         let (_, health_service) = tonic_health::server::health_reporter();
         let mut server_builder = Server::builder();
 
+        // by convention we add 1 to the API listen port and use that
+        // for the health check port.
         let port = port + 1;
         let addr = SocketAddrV4::new(addr, port);
         let server = server_builder
--- a/dataplane/api-server/src/lib.rs
+++ b/dataplane/api-server/src/lib.rs
@@ -16,7 +16,7 @@ use std::{
 
 use anyhow::{Context, Result};
 use aya::maps::{HashMap, MapData};
-use log::info;
+use log::{debug, info, error};
 use tonic::transport::{Certificate, Identity, Server, ServerTlsConfig};
 
 use backends::backends_server::BackendsServer;
@@ -31,6 +31,8 @@ pub async fn start(
     tcp_conns_map: HashMap<MapData, ClientKey, LoadBalancerMapping>,
     tls_config: Option<TLSConfig>,
 ) -> Result<()> {
+    debug!("starting api server on {}", addr);
+
     // Tonic itself doesn't provide a built-in mechanism for selectively
     // applying TLS based on routes, as TLS configuration is tied to the
     // entire server and managed at the transport layer, not at the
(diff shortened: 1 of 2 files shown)


dataplane/api-server/src/lib.rs
```diff
@@ -42,23 +44,37 @@ pub async fn start(
...
+        let port = port + 1;
+        let addr = SocketAddrV4::new(addr, port);
+        let server = server_builder
             .add_service(health_service)
-            .serve(SocketAddrV4::new(addr, port + 1).into())
-            .await
-            .unwrap();
+            .serve(addr.into());
+
+        debug!("gRPC Health Checking service listens on port {}", port);
```
shaneutt, line 54: As a personal preference, I would prefer to see this and the other line log the `{addr}:{port}` together, rather than logging the address, then the ports on separate lines. Thoughts?

dataplane/api-server/src/lib.rs
```diff
@@ -42,23 +44,37 @@ pub async fn start(
...
+        server.await.map_err(|e| {
+            error!("Failed serve gRPC Health Checking service, err: {:?}", e);
+            e
+        }).unwrap();
```
shaneutt, line 58: Since we're going to unwrap and panic anyway (meaning, we'll log the error, then we'll get a panic logged right after), I think we might as well just do a nice clear `.expect()` here with our error message.

dataplane/api-server/src/lib.rs
```diff
@@ -42,23 +44,37 @@ pub async fn start(
...
+
+        let tls_addr = SocketAddrV4::new(addr, port);
+        let tls_server = server_builder
             .add_service(BackendsServer::new(server))
-            .serve(SocketAddrV4::new(addr, port).into())
-            .await
-            .unwrap();
+            .serve(tls_addr.into());
+
+        debug!("TLS server listens on port {}", port);
```
shaneutt, line 73: Similar to above, what if we drop the debug at the top and just log `{addr}:{port}` together?

dataplane/api-server/src/lib.rs
```diff
@@ -42,23 +44,37 @@ pub async fn start(
...
+        tls_server.await.map_err(|e| {
+            error!("Failed to serve TLS, err: {:?}", e);
+            e
+        }).unwrap();
```
shaneutt, line 77: Similar to above, I think we can just put an `expect()` here :thinking:

dataplane/api-server/src/lib.rs
```diff
@@ -127,6 +143,9 @@ pub fn setup_tls(mut builder: Server, tls_config: &Option<TLSConfig>) -> Result<
             info!("gRPC mTLS enabled");
             Ok(builder)
         }
-        None => Ok(builder),
+        None => {
+            info!("gRPC TLS is not enabled");
```
shaneutt, line 147: :+1:

xtask/src/run.rs
```diff
@@ -45,7 +45,7 @@ fn build_dataplane(opts: &Options) -> Result<(), anyhow::Error> {
 }
 
 /// Build the controlplane
-fn build_contrlplane(opts: &Options) -> Result<(), anyhow::Error> {
+fn build_controlplane(opts: &Options) -> Result<(), anyhow::Error> {
```
shaneutt, line 48: whoops lol

shaneutt (changes requested):
Great! Thank you for taking the time to do this.

My comments are mostly minor things, take a look and let me know what you think.

buraksekili:
thanks for the review @shaneutt! i've updated the code accordingly

PR #360, opened 2025-02-04
Implement gateway class controller in Rust

### LINKED ISSUE #300: Rust ControlPlane: GatewayClass Controller
The purpose of this task is to create a `GatewayClass` controller in Rust. This controller is very simple, pretty much only marking the `GatewayClass` as `accepted` or not based on `ControllerName`.

See the Go version of this for reference: https://github.com/kubernetes-sigs/blixt/blob/archive/golang-control-plane/controllers/gatewayclass_controller.go

## TODO

- [x] Create a `GatewayClass` controller that marks managed `GatewayClass` resources as `accepted`
- [x] Update the `Gateway` controller to verify that any `GatewayClass` is accepted before programming routes

EandrewJones:
This PR implements the gateway class controller in Rust. 

It does the following:
- [x] Creates a `GatewayClass` controller that marks managed GatewayClass resources as accepted
- [x] Updates the `Gateway` controller to verify that any `GatewayClass` is accepted before programming routes
- [x] Defines `HasConditions` trait to enable the creation and usage of a generic `set_condition` util function across both our existing controllers.

To Do:
- [ ] Add integration tests for `GatewayClass` and Gateway which mirror our previous [Golang-based tests](https://github.com/kubernetes-sigs/blixt/blob/main/test/integration/gateway_test.go)

Closes #300.

### DIFF
--- a/controlplane/src/gateway_controller.rs
+++ b/controlplane/src/gateway_controller.rs
@@ -15,11 +15,13 @@ limitations under the License.
 */
 
 use futures::StreamExt;
+use log::debug;
 use std::{
     ops::Sub,
     sync::Arc,
     time::{Duration, Instant},
 };
+use utils::set_condition;
 
 use crate::*;
 use gateway_api::apis::standard::gateways::{Gateway, GatewayStatus};
@@ -74,6 +76,14 @@ pub async fn reconcile(gateway: Arc<Gateway>, ctx: Arc<Context>) -> Result<Actio
         "found a supported GatewayClass: {:?}",
         gateway_class.name_any()
     );
+    // Only reconcile the Gateway object if our GatewayClass has already been accepted
+    if !gatewayclass_utils::is_accepted(&gateway_class) {
+        debug!(
+            "GatewayClass {:?} not yet accepted",
+            gateway_class.name_any()
+        );
+        return Ok(Action::await_change());
+    }
 
     set_listener_status(&mut gw)?;
     let accepted_cond = get_accepted_condition(&gw);
@@ -127,11 +137,11 @@ pub async fn reconcile(gateway: Arc<Gateway>, ctx: Arc<Context>) -> Result<Actio
         let updated = update_service_for_gateway(gateway.as_ref(), &mut service)?;
         if updated {
             info!("drift detected; updating loadbalancer service");
-            let patch_parmas = PatchParams::default();
+            let patch_params = PatchParams::default();
             service_api
                 .patch(
                     val.name_any().as_str(),
-                    &patch_parmas,
+                    &patch_params,
                     &Patch::Strategic(&service),
                 )
                 .await
--- a/controlplane/src/gatewayclass_controller.rs
+++ b/controlplane/src/gatewayclass_controller.rs
@@ -1,0 +1,90 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+use futures::StreamExt;
+use std::{
+    ops::Sub,
+    sync::Arc,
+    time::{Duration, Instant},
+};
+
+use crate::*;
+use gateway_api::apis::standard::gatewayclasses::GatewayClass;
+use kube::{
+    api::{Api, ListParams},
+    runtime::{controller::Action, watcher::Config, Controller},
+};
+
+use gatewayclass_utils::*;
+use tracing::*;
+
+pub async fn reconcile(gateway_class: Arc<GatewayClass>, ctx: Arc<Context>) -> Result<Action> {
+    let start = Instant::now();
+    let client = ctx.client.clone();
+    let name = gateway_class
+        .metadata
+        .name
+        .clone()
+        .ok_or(Error::InvalidConfigError("invalid name".to_string()))?;
+
+    let gatewayclass_api = Api::<GatewayClass>::all(client);
+    let mut gwc = GatewayClass {
+        metadata: gateway_class.metadata.clone(),
+        spec: gateway_class.spec.clone(),
+        status: gateway_class.status.clone(),
+        // NOTE: Am I missing anything else here?
+    };
+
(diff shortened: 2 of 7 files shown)


EandrewJones:
@shaneutt @aryan9600 Looking for some light guidance to make sure this is on track before I go ahead with writing the integration tests.

controlplane/src/gatewayclass_controller.rs
```diff
@@ -0,0 +1,90 @@
...
+use tracing::*;
+
+pub async fn reconcile(gateway_class: Arc<GatewayClass>, ctx: Arc<Context>) -> Result<Action> {
+    let start = Instant::now();
+    let client = ctx.client.clone();
+    let name = gateway_class
+        .metadata
+        .name
+        .clone()
+        .ok_or(Error::InvalidConfigError("invalid name".to_string()))?;
```
shaneutt, line 41: I think perhaps it's not really "invalid" name, but if we could ever reach this error technically it was "no name provided"?

```suggestion
        .ok_or(Error::InvalidConfigError("no name provided for gatewayclass".to_string()))?;
```

FIXME

controlplane/src/gatewayclass_controller.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    let gatewayclass_api = Api::<GatewayClass>::all(client);
```
shaneutt, line 43:
```suggestion
```

controlplane/src/gatewayclass_controller.rs
```diff
@@ -0,0 +1,90 @@
...
+    let mut gwc = GatewayClass {
+        metadata: gateway_class.metadata.clone(),
+        spec: gateway_class.spec.clone(),
+        status: gateway_class.status.clone(),
+        // NOTE: Am I missing anything else here?
+    };
```
shaneutt, line 49: If the goal is to make a full clone, you should be able to do that with [Deref](https://doc.rust-lang.org/std/sync/struct.Arc.html#deref-behavior) and clone:
```suggestion
    let mut gwc = (*gateway_class).clone();
```

controlplane/src/gatewayclass_controller.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    if gateway_class.spec.controller_name != GATEWAY_CLASS_CONTROLLER_NAME {
+        // Skip reconciling because we don't manage this resource
+        // NOTE: May want to requeue in case this resource becomes relevant again in the
+        // future (e.g. the controllerName is changed to match ours in the event of typo,
+        // etc.)
+        return Ok(Action::requeue(Duration::from_secs(3600 / 2)));
```
shaneutt, line 56: We don't want to re-queue here as this could re-queue in an endless loop. We can rely on future updates to the gateway_class (including if the `controllerName` were to change for some reason) to re-enqueue the resource in our controller.

```suggestion
        return Ok(Action::await_change());
```

controlplane/src/gatewayclass_controller.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    if !is_accepted(&gateway_class) {
+        info!("marking gateway class {:?} as accepted", name);
+        accept(&mut gwc);
+        patch_status(&gatewayclass_api, name, &gwc.status.unwrap_or_default()).await?;
```
shaneutt, line 62:
```suggestion
        let gatewayclass_api = Api::<GatewayClass>::all(client);
        patch_status(&gatewayclass_api, name, &gwc.status.unwrap_or_default()).await?;
```

shaneutt (changes requested):
Took a first pass, looking good! Mostly minor comments.

controlplane/src/gateway_controller.rs
```diff
@@ -127,11 +137,11 @@ pub async fn reconcile(gateway: Arc<Gateway>, ctx: Arc<Context>) -> Result<Actio
         let updated = update_service_for_gateway(gateway.as_ref(), &mut service)?;
         if updated {
             info!("drift detected; updating loadbalancer service");
-            let patch_parmas = PatchParams::default();
+            let patch_params = PatchParams::default();
```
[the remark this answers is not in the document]
EandrewJones: I cannot tell you how often I make this stupid typo. It's hard coded into my fingers.

controlplane/src/gatewayclass_controller.rs
EandrewJones: This is indeed the aim, unless I can get away with not cloning at all. But we ultimately would need a thread-safe mutable reference (`Arc<Mutex<GatewayClass>>` to our gateway_class input then, but AFAICT from the kube.rs docs, `reconcile` will always be passed an `Arc<>` of the resource it's watching. Is that so?

controlplane/src/gateway_controller.rs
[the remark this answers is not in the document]
EandrewJones: Wait maybe it wasn' me. This was in the gateway controller.

controlplane/src/gateway_controller.rs
```diff
@@ -74,6 +76,14 @@ pub async fn reconcile(gateway: Arc<Gateway>, ctx: Arc<Context>) -> Result<Actio
         "found a supported GatewayClass: {:?}",
         gateway_class.name_any()
     );
+    // Only reconcile the Gateway object if our GatewayClass has already been accepted
+    if !gatewayclass_utils::is_accepted(&gateway_class) {
+        debug!(
+            "GatewayClass {:?} not yet accepted",
+            gateway_class.name_any()
+        );
+        return Ok(Action::await_change());
```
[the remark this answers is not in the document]
EandrewJones: Could we just re-enqueue here? Wouldn't that cause the gateway to keep re-enqueuing until it's GatewayClass was accepted?

That's not ideal because it's tantamount to naively long polling for changes to your GatewayClass that may never come versus the GatewayClass change driving the behavior.

Is there a an API for the watch mechanism you describe? If so, I can implement that.

controlplane/src/gatewayclass_controller.rs
shaneutt: It will always be an Arc, so if you want the inner of an arc you can do a deref clone (as shown above) OR an `as_ref().clone()`.

controlplane/src/gatewayclass_controller.rs
```diff
@@ -0,0 +1,87 @@
...
+        let gatewayclass_api = Api::<GatewayClass>::all(client);
+        patch_status(&gatewayclass_api, name, &gwc.status.unwrap_or_default()).await?;
+    }
+
+    let duration = Instant::now().sub(start);
+    info!("finished reconciling in {:?} ms", duration.as_millis());
+    Ok(Action::await_change())
+}
+
+pub async fn controller(ctx: Context) -> Result<()> {
```
nayihz, line 67: gatewayclass controller hasn't been started in main.rs.

EandrewJones: Good catch. TY.

Issue #357, opened 2025-01-28
Incorrect default kustomize manifests


buraksekili:
Besides the issues in the control plane container (with fixes being merged in #313, there are also minor issues in the current Kustomize files:

1.  **Outdated `kube-rbac-proxy` image in the control plane**

    -   The control plane uses the outdated `gcr.io/kubebuilder/kube-rbac-proxy:v0.13.0` image ([reference](https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr)).
    -   Considering that the control plane is now actively being developed in Rust, this proxy might no longer be necessary. These manifests looks like the leftovers from kubebuilder scaffolding.
2.  **Incorrect binary name in default manifests**
    -   The default manifests in `config/` point to an incorrect binary name https://github.com/kubernetes-sigs/blixt/blob/0858e9284dd1c6c9a46f81b24f9f7d8664c51290/config/manager/manager.yaml#L38
    -   The correct binary name should be `/workspace/manager`, (as per #313).
3.  **Incorrect ports in the data plane configuration**
    -   The data plane manifest specifies incorrect ports https://github.com/kubernetes-sigs/blixt/blob/0858e9284dd1c6c9a46f81b24f9f7d8664c51290/config/dataplane/dataplane.yaml#L39-L49
    -   The ports should be updated to `9875` instead of `9874`.

PR #364, opened 2025-02-13
default value update

kundan2707:
Fixes: #357


kundan2707:
Updated image location  based on https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/discontinue_usage_of_kube_rbac_proxy.md

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -15,7 +15,7 @@ spec:
           capabilities:
             drop:
               - "ALL"
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.0
+        image: quay.io/brancz/kube-rbac-proxy:v0.18.2
```
shaneutt, line 18: Where is this image coming from, specifically?

shaneutt (changes requested):

config/default/manager_auth_proxy_patch.yaml
shaneutt, line 18: Instead of pinning to what looks like a personally hosted image (though I see that's what's suggested in the documentation) I do think we should remove this file entirely. :thinking:

shaneutt (changes requested):

config/default/manager_auth_proxy_patch.yaml
kundan2707: @shaneutt  should i remove this file ?

shaneutt: Yes, let's remove it for now and if there end up being problems we can go "fix it forward" style. Thanks!

kundan2707: deleted and updated pull request

shaneutt (approved):
/approve
/lgtm

Issue #430, opened 2025-08-13
unify logging and dependency setup


hargut:
As of https://github.com/kubernetes-sigs/blixt/commit/e386d8031455e86b07efec0978f6c41aa6558460 the logging strategy for the project is not clear.

Some parts use `log::` and some other parts use `tracing::` and a few use both of the two:
https://github.com/kubernetes-sigs/blixt/blob/e386d8031455e86b07efec0978f6c41aa6558460/controlplane/src/gateway_controller.rs#L44-L45
https://github.com/kubernetes-sigs/blixt/blob/e386d8031455e86b07efec0978f6c41aa6558460/dataplane/api-server/src/lib.rs#L19
https://github.com/kubernetes-sigs/blixt/blob/e386d8031455e86b07efec0978f6c41aa6558460/controlplane/src/main.rs#L23

# Suggestion
- Use `tracing::` for all logs.
- Unify dependencies through workspace dependencies.

This will for future use aslo allow to trace connection between the `dataplane` and the `controlplane`.

PR #434, opened 2025-08-13
fix(logs): remove unused env_logger

hargut:
after switching to tracing this is no longer required

fixes ci issue on main
related #430


shaneutt (approved):
/approve
/lgtm

PR #441, opened 2025-08-29
feat(loader): eBPF program and map pinning

### LINKED ISSUE #426: dataplane configuration and connections are lost during redeployment
# Problem
A redeployment of the `dataplane` causes a full reload of the ebpf program and re-initialization of the configuration and connection maps.

As of now some k8s resources need to be recreated in order to reconfigure the `dataplane`. L4 loadbalancing is interrupted and the hosts active TCP connections are dropped during a `dataplane` redeployment.

# Possible Solutions
- pin ebpf programs and maps to avoid dropping all configuration and connection during redeployments
  - provide a cli option to reload the ebpf programs and maps (e.g. during development use)
  - always reload the ebpf programs in case the `dataplane` version changed
  - default to use pinned ebpf programs and maps
  - this option can keep loadbalancing up and active TCP connections alive
- update the `controlplane` logic to fully reconfigure a new `dataplane` pod
   - interrupts L4 load-balancing and drops nodes active TCP connections without the first option
   - this will be required during node scale-outs on external load balancer setups

hargut:
mount bpf fs in kind cluster during tests
and in dataplane pod

fixes #426

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1304,6 +1304,7 @@ dependencies = [
  "aya-log",
  "clap",
  "common",
+ "thiserror 2.0.14",
  "tokio",
  "tracing",
  "tracing-log",
--- a/config/dataplane/dataplane.yaml
+++ b/config/dataplane/dataplane.yaml
@@ -17,13 +17,24 @@ spec:
         app: blixt
         component: dataplane
     spec:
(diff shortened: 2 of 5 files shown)


shaneutt (approved):
/approve
/lgtm

PR #443, opened 2025-08-29
feat(controlplane): gateway resource cleanup

### LINKED ISSUE #440: gateway controller finalizers are correctly writtten, resoruces cleanup on deletion incomplete
As of now there is code that apparently should set the finalizers field, but they never appear on the k8s object.

Further on gateway deletion not all created sub-resources are cleaned up.

hargut:
use a finalizer to clean up created resources

fixes #440


shaneutt (approved):
/approve
/lgtm

